PR #4, opened 2018-10-27
Fix the examples

davidkretch:


adambanker:
Hey David, some of the updates we made to publish the package make the start_here script not work as expected. The two issues I see are:

1. The "aws-sdk-js/apis" directory no longer is included with the package. We should write instructions on how one could get that info so that they can build the packages.

2. The configure file gives me permission errors again. We just need to update the permissions on the file and update that.

davidkretch:
Yeah, I think we might include aws-sdk-js as a submodule. Also, the start_here.R script is going away entirely. 

I don't think I can update the permissions on Windows. I will see what I can do though.

adambanker:
The examples work so I am going to merge the pull request.

PR #26, opened 2018-11-12
REST XML handlers and S3 support

adambanker:


make.paws/R/xmlutil.R
```diff
@@ -61,7 +196,13 @@ xml_parse_structure <- function(node, interface) {
     if (length(children) == 0) {
       # TODO: Implement.
     }
+    
+    # if (flattened) {
```
davidkretch: note to self: check

davidkretch:
We should rebase this branch. Let's check in tomorrow.

make.paws/R/xmlutil.R
davidkretch: Looks ok.

davidkretch:
This PR fails its tests and I don't know why. I am going to move all the changes unrelated to REST XML to a new PR.

make.paws/R/net.R
```diff
@@ -100,11 +100,14 @@ issue <- function(http_request) {
     body = body
   )
 
+  response_body <- httr::content(r, as = "text", encoding = "UTF-8")
+  if (is.na(response_body)) response_body <- httr::content(r, as = "raw")
```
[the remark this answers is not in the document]
davidkretch: The body is now kept as raw and decoded as appropriate in the `unmarshal` functions.

make.paws/R/make_service.R
```diff
@@ -2,6 +2,7 @@
 make_service <- function(api) {
   protocol <- protocol_package(api)
   signature <- api$metadata$signatureVersion
+  if (signature == "s3") signature <- "v4"
```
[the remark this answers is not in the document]
davidkretch: There is now an `s3_sign_request_handler` which points to the V4 signer.

adambanker:
I believe this is ready to review. There is one thing, where for the list of structures test, test_handlers_restxml.R line 434, I cannot properly populate a list of multiple structures. That isn't a restxml issue, and I don't know if it is actually an important test. The test works fine on a list of one structure.

make.paws/tests/testthat/test_handlers_restxml.R
```diff
@@ -0,0 +1,692 @@
...
+
+#-------------------------------------------------------------------------------
+
+# XML tests
+
+test_that("Basic XML Case1", {
+  op_test <- Operation(name = "OperationName")
+  op_input_test <- function(OperationRequest) {
+    args <- list(OperationRequest = OperationRequest)
+    interface <- structure(list(
```
davidkretch: This should use the functions in populateutil, e.g. `Structure()`.

make.paws/tests/testthat/test_handlers_restxml.R
```diff
@@ -0,0 +1,692 @@
...
+  )))
+})
+
+test_that("List of Structures Case1", {
+  op_test <- Operation(name = "OperationName")
+  op_input_test <- function(OperationRequest) {
+    args <- list(OperationRequest = OperationRequest)
+    interface <- structure(list(
+      OperationRequest = structure(list(
+          ListParam = structure(list(
```
davidkretch: A list of structures should look like the following. The list shouldn't have any named elements itself.
```r
list(
  ListParam = list(
    list(
      Foo = 1,
      Bar = 2
    )
  )
)
```

make.paws/R/xmlutil.R
```diff
@@ -1,16 +1,150 @@
...
+
+# Convert an XML string to an R list.
+xml_to_list <- function(value) {
+  if (is_empty(value)) return(NULL)
+  result <- xml2::as_list(xml2::read_xml(value))
+  return(result)
+}
+
+# Convert list to XML text
+xml_list_to_character <- function(value) {
```
davidkretch: 1. This should probably be `list_to_xml` since it's doing the reverse of `xml_to_list`.
2. It inserts newlines and an extra xml tag, which you can get rid of with `as.character(xml, options = "no_declaration")`. See [write_xml](https://www.rdocumentation.org/packages/xml2/versions/1.2.0/topics/write_xml).

make.paws/tests/testthat/test_handlers_restxml.R
```diff
@@ -0,0 +1,692 @@
...
+
+  input <- op_input_test(
+      OperationRequest = list(
+      Description = "bar",
+      Name = "foo")
+  )
+  req <- new_request(svc, op_test, input, NULL)
+  req <- build(req)
+  r <- req$body
+  expect_equal(r, xml_list_to_character(list(
```
davidkretch: I think we should be copy and compare against the same output that the Go SDK test is comparing against, or at least a minimally-modified version of that. This compares against a tailored object put through a function that is also used by the function being tested.

make.paws/tests/testthat/test_handlers_restxml.R
```diff
@@ -0,0 +1,692 @@
...
+})
+
+
+#-------------------------------------------------------------------------------
+
+# XML tests
+
+test_that("Basic XML Case1", {
+  op_test <- Operation(name = "OperationName")
+  op_input_test <- function(OperationRequest) {
```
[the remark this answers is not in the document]
adambanker: Hmmmm. Looking at line 77 [https://github.com/aws/aws-sdk-go/blob/master/private/protocol/xml/xmlutil/build.go](url)
I'm wondering if the more suitable interface would be:
```
interface <- Structure(
      _ = Structure(
          Description = Scalar(type = "string"),
          Name = Scalar(type = "string"),
          .attrs = list(locationName = "OperationRequest", xmlURI = "https://foo/")
     )
)
```
Which then gets back to the underscore issue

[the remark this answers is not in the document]
davidkretch: That is feasible but at least right now, the tags like locationName that would be attached to `shape["_"]` are currently attached to `shape`. Also, that would change the calling signature of the operation -- it would require your call to the operation be `operation(_ = list(Description = "foo", Name = "bar"))`.

make.paws/R/xmlutil.R
adambanker: addressed in last commit

make.paws/R/xmlutil.R
```diff
@@ -1,16 +1,150 @@
...
+  return(result)
+}
+
+xml_build_structure <- function(params) {
+  result <- list()
+  for (name in names(params)) {
+    child <- params[[name]]
+
+    if (get_tag(child, "locationName") == "") {
+      attr(child, "locationName") <- name
```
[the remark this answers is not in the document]
adambanker: addressed in previous commit

make.paws/tests/testthat/test_handlers_restxml.R
adambanker: addressed in last commit. I also added in xml namespace support

adambanker:
@davidkretch I believe I have now addressed all of your comments.

make.paws/R/convert.R
```diff
@@ -73,3 +73,10 @@ timestamp_format <- function(name) {
   )
   return(format_string)
 }
+
+convert_xml_boolean <- function(value) {
+  if (value) {
+    return("true")
+  }
+  return("false")
+}
```
davidkretch: Can we use `convert_boolean` instead?

make.paws/R/xmlutil.R
```diff
@@ -1,16 +1,188 @@
...
+}
+
+# Convert an XML string to an R list.
+xml_to_list <- function(value) {
+  if (is_empty(value)) return(NULL)
+  result <- xml2::as_list(xml2::read_xml(value))
+  return(result)
+}
+
+add_xmlns <- function(xml_list, xmlns = "") {
```
davidkretch: Move below `list_to_xml`.

make.paws/R/xmlutil.R
```diff
@@ -1,16 +1,188 @@
...
+  } else {
+    body_xml <- ""
+  }
+
+  request$body <- body_xml
+  request$http_request$body <- body_xml
+  return(request)
+}
+
+xml_build <- function(params) {
```
davidkretch: Move up to be the first `xml_build` function.

make.paws/R/xmlutil.R
```diff
@@ -1,16 +1,188 @@
...
+        result <- c(result, parsed)
+      } else{
+        result[[location_name]] <- parsed
+      }
+    }
+  }
+  return(result)
+}
+
+xml_build_structure_no_names <- function(params) {
```
davidkretch: When does a structure have no names? To my knowledge it should always have names.

make.paws/R/xmlutil.R
```diff
@@ -1,16 +1,188 @@
...
+  params <- request$params
+
+  if (is_empty(params)) {
+    body_xml <- ""
+    request$body <- body_xml
+    request$http_request$body <- body_xml
+    return(request)
+  }
+
+  locationName <- get_tag(params, "locationName")
```
davidkretch: Change to snake case.

adambanker:
@davidkretch I am done making updates.

adambanker:
@davidkretch should i make those changes you requested now?

davidkretch:
@adambanker I was making them. I noticed that `add_xmlns` also needs to use `add_tags` and was changing that and adding tests.

davidkretch:
I'm going to look into the structure names issue but then I think this is ready to merge.

make.paws/tests/testthat/test_handlers_restxml.R
```diff
@@ -0,0 +1,664 @@
...
+})
+
+test_that("List of Structures Case1", {
+  op_test <- Operation(name = "OperationName")
+  op_input_test <- function(ListParam) {
+    args <- list(ListParam = ListParam)
+    interface <- Structure(
+        ListParam = List(
+          Structure(
+            Scalar(type = "string", .attrs = list(locationName = "value"))
```
davidkretch: the issue with the missing names is this needs a name, e.g. `Element = Scalar(...)`.

make.paws/tests/testthat/test_handlers_restxml.R
```diff
@@ -0,0 +1,664 @@
...
+        .attrs = list(locationName = "OperationRequest",
+                      xmlURI = "https://foo/")
+    )
+    return(populate(args, interface))
+  }
+
+  input <- op_input_test(
+      ListParam = list(
+        list(
+          value = list("one")
```
davidkretch: and these should probably be only `"one"`, etc.

davidkretch:
@adambanker Can you look at the latest two comments on the structure names. Structures should have names but currently it isn't enforced. I don't think `xml_build_structure` needs to handle the case that there aren't any names. We should eventually make it so that Structure requires names.

make.paws/tests/testthat/test_handlers_restxml.R
adambanker: I made that update and updated build_structure

make.paws/tests/testthat/test_handlers_restxml.R
adambanker: It expected an input in the form of
```
input <- op_input_test(
      ListParam = list(
          list(Element = "one"),
          ...
```

davidkretch: Ah right. Yes.

davidkretch:
Looks good to me.

PR #28, opened 2018-11-14
Add an install option in the makefile

davidkretch:
* Add an option in the makefile to build and install a package, e.g. `make install-paws.s3`.
* Automatically update the package version when the package has been updated, as represented by uncommitted changes.
* Change the makefile so you supply it with a package name, rather than an API file name. For example to build a package now you type `make paws.ec2`.
* Move some tests to new test files since the functions' files had also changed.
* Delete `parse_file_name` which was no longer used anywhere.
* Add placeholder functions for REST XML handlers and V2 and S3 signers. Without these, packages that rely on them will fail to build.


adambanker:
This doesn't seem to work for me. When I run `make paws.ec2` I get:
```
build paws.ec2
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
[1] "paws.acm"
```

davidkretch:
Thanks. I have updated the make recipe to avoid the `-P` option which was no available. Please try again.

adambanker:
Can you get rid of this message when I do make package:
```
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```

davidkretch:
I made it use `grep -e <pattern> <file>`. I'm guessing that is what your grep requires.

davidkretch:
Does the following command work for you

```bash
grep -e "^paws.ec2\b" PACKAGES.txt
```

adambanker:
Yes that works for me

davidkretch:
I added the quiet option. Maybe that will suppress that output.

adambanker:
Now make paws.ec2 gives me this output again:
```
build paws.ec2
[1] "paws.acm"
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```

davidkretch:
Does `grep -e` give no output?

adambanker:
When I run this I get:
```
> grep -e "^paws.ec2\b" PACKAGES.txt

paws.ec2	ec2

```

davidkretch:
Ok it looks like `grep -e` alone is ok. I will remove change from the last commit.

adambanker:
Ok, for Mac, I needed to install the gnu versions of grep and sed. I did so with the following commands:
```
brew install grep --with-default-names
brew install gnu-sed --with-default-names
```
Now I am getting this error, though:
```
sed: -e expression #1, char 36: unterminated `s' command
sed: -e expression #1, char 34: unterminated `s' command
```

davidkretch:
Ohhh, the issue is not in the makefile anymore, it is in `script/update_version.sh`.

davidkretch:
What happens when you run `sed -i s/^paws.ec2\t.*/paws.ec2\t0.0.0.9004/ VERSIONS.txt`

adambanker:
It doesn't return anything

davidkretch:
At least it doesn't fail. What does `grep -e "paws.ec2" VERSIONS.txt" return?

adambanker:
```
paws.ec2	0.0.0.9000
paws.ec2	0.1.0
paws.ec2	0.1.0
paws.ec2	0.1.0
paws.ec2	0.1.0
paws.ec2	0.1.0
paws.ec2	0.1.0
paws.ec2	0.1.0
paws.ec2	0.1.0
```

davidkretch:
Haha, hilariously broken.

davidkretch:
It is due to the repeated paws.ec2 entries in VERSIONS.txt. Can you run `git checkout HEAD VERSIONS.txt`. I will look at explicitly checking for repeated entries.

PR #40, opened 2018-12-01
Add do not edit warning to generated files

adambanker:


make.paws/R/make_package.R
```diff
@@ -98,7 +101,9 @@ copy_customizations <- function(api, path) {
   if (length(src_file) > 0) {
     package <- package_name(api)
     filename <- sprintf("%s_customizations.R", package)
-    file.copy(from = src_file, to = file.path(path, filename))
+    warning_text <- unlist(edit_warning(list(), new_line = T))
+    cat(warning_text, file = file.path(path, filename))
```
davidkretch: Can you change this to `write` or `writeChar` instead of `cat`? I feel one of these would be clearer.

davidkretch:
A couple requests:
1. Can you change `edit_warning` to `add_edit_warning`?
2. Can you change the text to "This file is generated by make.paws. Please do not edit here."
3. See other comment.

make.paws/R/make_package.R
adambanker: I changed it to write

davidkretch:
Looks good to me.

PR #47, opened 2018-12-08
Add the correct packages to Imports

davidkretch:
* Get the packages used by each Paws package to the `DESCRIPTION` file's Imports section from a file, rather than from `make.paws`'s `DESCRIPTION`.
* Generate the list of packages by scanning the code for the package qualifier `::` and save the list in `extdata/imports.csv`.
* Make `configure.win` and `cleanup.win` run the regular `configure` and `cleanup`.


davidkretch:
It isn't, but it apparently only runs the `cleanup` script under special conditions. See [Configure and cleanup](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Configure-and-cleanup). It would make sense to delete the contents at the beginning of `configure`.

davidkretch:
No I don't have any more. Thanks

Issue #35, opened 2018-11-28
Ensure that output objects don't have extra attributes


davidkretch:
e.g. the attributes we add to help unmarshal API responses.

davidkretch:
Fixed by #51.

PR #51, opened 2018-12-14
Clean output objects

davidkretch:
* Remove references to underscore elements `_` which we do not have. (#41)
* Change how tags are implemented and change tag function names:
    + Store all tags as a list in the `tags` attribute, which makes it easier to delete.
    + Rename all tag function to `tag_...`, mostly in order to avoid namespace collisions with API operations like `add_tags`. Rename `type.R` to `tags.R`.
    + Change the `populateutil` functions to work with the new tags.
* In `send_request`, delete all tags before returning API response objects. (#35)


adambanker:
@davidkretch this works for me.

One unrelated thing I noticed when testing, is that when I do `paws.batch::describe_job_definitions()`, I still see old job definitions from the example that I deregistered. I cannot see these job definitions in the aws online console though.

davidkretch:
@adambanker Thanks, I'll take a look. I think there might be some things exposed to the API that are not in the console.

PR #52, opened 2018-12-15
WIP: updates query error parsing

adambanker:


davidkretch:
I created that, but iirc it used an actual API response.

adambanker:
Hmmm, none of the response I have received look like that - which would also explain why it wasn't working. I will update it and add some tests if that is ok with you @davidkretch

adambanker:
This is ready to be reviewed. While I see where the ErrorResponse tag came from in their unmarshal error code, the actual responses look like as follows:
[S3 Error Responses](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html)

davidkretch:
@adambanker I think this looks good with the exception of a couple comments. I also tried it successfully with ACM (query), EC2 (ec2query), and S3.

adambanker:
@davidkretch are you going to add comments?

davidkretch:
@adambanker Sorry, I meant the line-item comments above.

adambanker:
Can you remake those comments? I don't see them

adambanker:
@davidkretch i made the updates

davidkretch:
Looks good to me.

PR #55, opened 2018-12-18
creates a user agent header

adambanker:


make.paws/R/handlers_core.R
```diff
@@ -1,6 +1,16 @@
 # Add the name/version to the User-Agent request header.
 # TODO: Implement.
 sdk_version_user_agent_handler <- function(request) {
+  paws_version <- packageVersion(sprintf("paws.%s",
```
davidkretch: This should use `methods::getPackageName()`. I don't think `request$client_info$service_name` is guaranteed to work.

davidkretch: In the sense that I think the package name is actually kind of complicated and not necessarily the service name.

davidkretch: Also I think it needs to be `utils::packageVersion` -- I think R CMD CHECK will throw errors if it isn't specified.

davidkretch:
AppVeyor failed due to some error in their environment. Could you try re-running it.

davidkretch:
Looks good to me.

PR #56, opened 2018-12-20
Add usage documentation

davidkretch:
* Add a usage section to each function's documentation, showing all possible parameters and example values.
* Reorganize `make_docs.R` to put the `make_docs_...` functions at the top.
* Store enum values in the interfaces, and no longer store the shape name. We do not use shape name in the SDK packages, only when generating the packages.


adambanker:
This looks good to me.

PR #72, opened 2019-01-27
Move common functions to package paws.common

davidkretch:
* Move common functions out of codegen and each generated package into R package `paws.common`.


adambanker:
@davidkretch, can you explain why the generated packages still copy a lot of the code? For instance, if I run, `make install-paws.ec2` the result still has handlers like handler_restjson.R in it.

davidkretch:
make.paws isn’t updated. Its tests will fail until paws.common is in master. Otherwise it will try to fetch paws.common and fail.

adambanker:
So are the generated packages not using paws.common then?

davidkretch:
Right, they won’t use them until make.paws is also updated.

PR #75, opened 2019-02-03
Don't merge: Moves usage to examples in documentation

adambanker:


davidkretch:
After messing with Roxygen in the paws.common branch, you can create a new section with `@section Whatever:` and make it pre-formatted with triple backticks. See https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html.

adambanker:
@davidkretch , I'll try that out and if I can get that to work, I think I'll just make a new pull request

PR #77, opened 2019-02-07
moves usage to an additional parameters section

adambanker:


davidkretch:
Looks good. I will also try it out tonight.

PR #78, opened 2019-02-07
Adds donttest to examples so it can pass r cmd check

adambanker:


adambanker:
It currently has warnings when running R CMD check because it tries and fails to run the integration tests. I am not sure if rectifying that should be part of this merge request.

davidkretch:
Looks good to me. Let's deal with that in another PR.

PR #80, opened 2019-02-08
Fix R CMD check issues

davidkretch:
* Don't generate an examples section when empty.
* Don't copy the license to each package (can't have it in the package directory).
* Make package Depends on R version 3.4.0, instead of 3.5.1 (must depend on an x.x.0 release).
* Add testthat to Suggests.
* Remove links before creating help page titles; mask square brackets in titles so they don't look like markdown links.


adambanker:
@davidkretch looks good to me.

PR #81, opened 2019-02-09
Delete testthat.R file

davidkretch:
Delete all code for making the testthat.R file.

We actually don't need testthat.R for the command in our makefile to run a package's integration tests. It works without any changes.


adambanker:
It works!

PR #82, opened 2019-02-09
Handle missing operation descriptions

davidkretch:
Previously, `html_to_text` would fail if run on `NULL`, as when an operation had no description. Return `""` when the input is `NULL`.


davidkretch:
@adambanker This is ready -- codegen runs through all packages without errors.

adambanker:
Cool. I checked it and it worked.

PR #83, opened 2019-02-09
Regenerates the packages to use paws.common

adambanker:


davidkretch:
I think we have different versions of aws-sdk-js, since e.g. `paws.amplify` wasn't updated. I think this means that `make deps` should be a dependency of `make build`, so it gets the committed version of aws-sdk-js.

adambanker:
@davidkretch i updated the dependencies and am rerunning make build now

davidkretch:
I fixed VERSIONS.txt which had inconsistent version numbers -- some still were at 0.0.2. There's probably a bug in `script/update_version.sh`.

I also committed the updated aws-sdk-js so now the version you get with `make deps` is the version that the packages were generated from.

adambanker:
@davidkretch this works for me. Are you ok with me merging it?

davidkretch:
Looks good to me too. I got it.

PR #84, opened 2019-02-10
Fix misc CRAN submission issues

davidkretch:
* The description cannot start with the package name.
* Remove typos and acronyms (e.g. "AWS" and "SDK") that get flagged as notes.
* Remove the Remotes section, which CRAN packages cannot have. Move the GitHub repo for `covr` to the Travis config file.
* Use `httpbin.org` to test network resources, like the `httr` package does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: r
 
 cache: packages
 
+r_github_packages:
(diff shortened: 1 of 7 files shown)


adambanker:
Works for me

PR #94, opened 2019-03-04
Make a single package supporting all AWS services

davidkretch:
* Example usage of the new package.
```r
ec2 <- paws::ec2
ec2$describe_instances()
?ec2
?ec2_describe_instances
```
---
* Each service now lives in two objects, defined in `<service>_service.R`.
    * A public `service` object containing all operations, e.g. `ec2`. This object is exported.
    * A private `.service` object containing the service client, config, handlers, and interfaces, e.g. `.ec2`. This object is not exported, and is not ordinarily accessible by the user. To modify a service's handlers, you would use `.service$handlers$build <- foo`.
---
* Code generation now uses a simpler API with `template()` and `render()` and each template generally represents a more complete part of the final code.
    * `template()` stores a template. It takes a string or code and returns a string.
    * `render()` interpolates any variables in the template and returns a string.
    * Alternatives/old ways used R code objects or `glue`ed strings, which have limits: R code objects can't capture comments, and for strings meant to be `glue`ed, text within `{}` gets interpreted as code and can fail.
    * Example usage:
```r
templ <- template(
  `
  # This is a comment.
  foo <- ${bar}
  `
)
render(templ, bar = 123)

# Returns:
"# This is a comment
foo <- 123"
```
---
* Rename files.
    * In general `make_foo.R` -> `foo.R`.
    * Split general text functions out of `make_docs.R` into `text.R` and kept the rest in `docs.R`.
    * `make_package.R` is split into `make_sdk.R` which makes the package skeleton, and `process_api.R` which makes the files for a single API at a time.
---
* Update makefile and now rely on sole exported function `make_sdk()`.
    * Add description
---
* Update `tag_del()` in `paws.common` to take an optional `tags` argument, which when provided will recursively delete all such tags in a tagged object.


adambanker:
@davidkretch if I do `?paws::ec2` it would be nice to get links to the documentation for all the ec2 functions. Do you want to add that in a separate pull request?

davidkretch:
Yeah that is the plan but I was going to do it later. I can do it on this branch but it might be a few more days.

adambanker:
@davidkretch this works for me in its current state. We should make sure the documentation is fully updated before we submit to CRAN, but I am sure you already had that same plan. I am going to merge it.

PR #99, opened 2019-03-11
Updates examples to use the singular paws package

adambanker:


davidkretch:
What are your thoughts on pulling out the service objects from the `paws::` namespace e.g...

```
ec2 <- paws::ec2
ec2$whatever
```

adambanker:
@davidkretch we shouldn't merge this until we update the repo to use the singular paws package.

davidkretch:
Oh right.

adambanker:
> What are your thoughts on pulling out the service objects from the `paws::` namespace e.g...
> 
> ```
> ec2 <- paws::ec2
> ec2$whatever
> ```

I am leaning towards doing `paws::ec2` just because it makes it explicit what comes from the paws package. The way you are suggesting makes it look more python-like, but I don't know if that is something we care about.

davidkretch:
Yeah, I'm ok either way, but I think there are two things that could affect us -- I think it's somewhat more readable and I expect would be the popular way of doing it, and also assuming we ever get custom configuration working, I think it might need that syntax to modify this particular ec2 object rather than the original one that lives in the paws namespace.

adambanker:
Interesting. One thing I like about doing it `paws::ec2` is that it makes it easy to look up the documentation by doing `?paws::ec2`.

You can currently do `?ec2` and it would work, but the following examples would not bring up the proper help documentation:
```
s3 <- paws::ec2

# Brings up the s3 documentation instead of ec2
?s3

my_ec2 <- paws::ec2
# Brings up no documentation message
?my_ec2
```

davidkretch:
Ha, that's a funny example.

davidkretch:
Actually, we should have more about getting help: The package manual lists all the available services, which you can access with `help(package = "paws")`. Each service's manual page lists all available operations, etc.

README.md
```diff
@@ -27,13 +27,13 @@ JavaScript](https://github.com/aws/aws-sdk-js).
 
 ### Installation
 
-Paws is currently available on GitHub. Each AWS service has its own R
-package within Paws. For example, to use EC2, use the `paws.ec2` package.
-See [the list of available packages here](service). Install a package
-(e.g. `paws.ec2`) using:
+Paws is currently available on GitHub. Each AWS service is available
+within Paws. For example, to use EC2, use `paws::ec2`.
+See [the list of available packages here](service). Install paws
```
davidkretch: This folder will disappear.

README.md
```diff
@@ -27,13 +27,13 @@ JavaScript](https://github.com/aws/aws-sdk-js).
...
+(e.g. `paws`) using:
```
davidkretch: No longer need "(e.g. `paws`)".

PR #103, opened 2019-03-15
Fix Paws CRAN notes/warnings and other issues

davidkretch:
* If an operation has no documentation, make the operation name into the documentation title. This fixes warnings for operations that were missing titles.

* Fix the "unknown macro" warnings by changing "\\uxxxx" to "`U+xxxx`" and control characters e.g. "\n" to "`\\n`".

* Fix the service descriptions, which were cut off after the first line.

* Break long lines in the usage section to avoid R CMD check warnings.

* Make the service objects (e.g. `ec2`) into functions. Example usage: `ec2 <- paws::ec2()`.

    This is in anticipation of the future Paws API, in which you would set custom configuration for each API (e.g. change the region, endpoint, etc.) by adding arguments to the function.

    Currently the service functions accept no arguments. This means that code written against this API will run in future versions that accept some arguments, but code written against future versions won't run in this version, which I think is desireable.

### DIFF
--- a/make.paws/R/docs.R
+++ b/make.paws/R/docs.R
@@ -233,7 +233,7 @@ escape_special_chars <- function(text) {
   # Unicode character codes: \\uxxxx to `U+xxxx`
   result <- gsub("\\\\\\\\u([0-9a-fA-F]{4})", "`U+\\1`", result)
 
-  # Escape character codes: e.g. \n to `\\n`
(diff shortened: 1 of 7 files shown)


davidkretch:
Travis CI for R is currently down.

adambanker:
Looks good.

PR #108, opened 2019-03-17
Move customizations to paws.common

davidkretch:
Automatically add customizations to the handlers for a given service by looking them up by name from the new `customizations` object.

This avoids having to export more functions to paws to get the customizations to work. It also means that fixes to customizations get distributed with paws.common which is smaller and easier to update.

This merge request also removes code related to customizations from make.paws, and no longer exports handler setting functions from paws.common, which were only exported to support customizations that are now internal.


davidkretch:
@adambanker Travis is still broken.

PR #178, opened 2019-08-18
paws.common CRAN submission

davidkretch:


davidkretch:
All builds are failing on AppVeyor until CRAN is updated with the latest paws.common.

PR #184, opened 2019-08-25
Adds the ability to specify custom endpoints

adambanker:


make.paws/R/service.R
```diff
@@ -85,6 +92,36 @@ service_description <- function(api) {
...
+  paste("@param", argument, desc, sep = "\n")
+}
+
+# Return the Service Syntax.
+service_syntax <- function(api) {
+  section <- "@section Service Syntax:"
+  service <- package_name(api)
+  syntax <- sprintf(
+    '```
+    Svc <- %s(
```
davidkretch: This should be 

```
svc <- %s(
    config = list(
        endpoint = "string"
    )
)
```

paws.common/R/client.R
```diff
@@ -98,11 +101,11 @@ resolver_endpoint <- function(service, region, endpoints, scheme = "https") {
...
 
 # client_config returns a ClientConfig configured for the service.
-client_config <- function(service_name, endpoints) {
-  s <- new_session()
+client_config <- function(service_name, endpoints, cfgs = NULL) {
+  s <- new_session(cfgs)
   region <- s$config$region
   if (s$config$endpoint != "") {
-    endpoint <- sprintf("https://%s", s$config$endpoint)
+    endpoint <-  s$config$endpoint
```
davidkretch: This should add https if it isn't already in the endpoint string.

make.paws/R/service.R
```diff
@@ -85,6 +92,36 @@ service_description <- function(api) {
...
+  argument <- comment(paste(argument, collapse = "\n"), "#'")
+  desc <- "An optional list of custom configurations for the service. Currently
+           supports adding a custom endpoint."
+  desc <- comment(paste(desc, collapse = "\n"), "#'")
+  paste("@param", argument, desc, sep = "\n")
+}
+
+# Return the Service Syntax.
+service_syntax <- function(api) {
+  section <- "@section Service Syntax:"
```
davidkretch: Should be "Service syntax"

paws.common/R/client.R
adambanker: The Go SDK and other SDKs have you specify the full endpoint with https:
`svc := ec2.New(sess, &aws.Config{Endpoint: aws.String("https://test.us-west-2.amazonaws.com")})`

davidkretch: Ok, sounds good.

adambanker:
@davidkretch you can now set credentials, region, and endpoint in a custom config. So you could do the Google Cloud Storage example without separately setting environment variables

adambanker:
@davidkretch it now supports using temporary security credentials. I think it is in good shape for you to review now even though some checks failed because the updates led to  a decrease in code coverage. Once you review it and we can discuss the approach I took, I can add in more tests to ensure we are covering edge cases/increase code coverage.

davidkretch:
I'm ok with most of this but one big issue, which is that I don't think assuming a role via STS can be a  provider in the credential provider chain, because it is independent of the other method of getting credentials.  You should be able to assume a role after first using environment variables, IAM role, etc.  [This page](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/) has an example of assuming a role, where your initial credentials could be any method.

I may be misunderstanding the implementation, but I don't think we can include it in the provider chain by merging credentials together.  Since the provider chain should return with the first set of credentials it finds, I don't think this can be used to first get credentials to assume the role, then merge in the assumed role credentials.  Also, if we get incomplete credentials from one of the providers, I don't think we should ever be merging with another provider's credentials.

I have some other minor issues that I'll add as line item comments.

paws.common/R/client.R
```diff
@@ -64,10 +64,65 @@ Client <- struct(
...
-# new_session returns a Session with default configuration.
-new_session <- function() {
+# Populate config
+merge_in_config <- function(dst, other){
+  if (is.null(other)){
+    return(dst)
+  }
+  # Populate credentials
+  if (!is.null(other$credentials)){
+    if (!is.null(other$credentials$creds$access_key_id)){
```
davidkretch: I think if creds is provided, the existing one should be entirely overwritten.

paws.common/R/client.R
```diff
@@ -64,10 +64,65 @@ Client <- struct(
...
+  # Fill in Config with any user defined values
+  cfg <- merge_in_config(cfg, cfgs)
+  # If region not defined, set it
+  if (nchar(cfg$region) == 0) {
+    cfg$region <- get_region()
+  }
+  # If credentials passed in, use those credentials
+  if (!is.null(cfgs) && !is.null(cfgs$credentials)){
+    cfg$credentials$provider <- c(
+      eval(
```
davidkretch: Can this be `function() cfg$credentials$creds`?

paws.common/R/config.R
```diff
@@ -95,6 +98,41 @@ check_config_file_region <- function() {
...
+    config_values <- ini::read.ini(config_path)
+    if (is.null(config_values[[aws_profile]])) return(NULL)
+  }
+
+  role_arn <- config_values[[aws_profile]]$role_arn
+  source_profile <- config_values[[aws_profile]]$source_profile
+  role_session_name <- config_values[[aws_profile]]$role_session_name
+  
+  if (is.null(role_session_name)) {
+    role_session_name <- paste0(sample(LETTERS, 15, replace = TRUE),
```
davidkretch: This needs a comment -- not sure why this is random like this.

paws.common/R/config.R
```diff
@@ -95,6 +98,41 @@ check_config_file_region <- function() {
   return(region)
 }
 
+# Tries to get the region from the config file
+check_config_file_assume_role <- function() {
```
davidkretch: Probably should check whether it got a complete set of creds before returning.

paws.common/R/credentials.R
```diff
@@ -14,18 +14,60 @@ Credentials <- struct(
...
 # Get credentials using the first set of credentials found by the list of
 # credential providers in creds$provider. If no credentials are found, return
 # the original credentials object.
 get_credentials <- function(credentials) {
+  creds <- Credentials()$creds
   for (provider in credentials$provider) {
-    creds <- provider()
-    if (!is.null(creds)) {
+    provider_creds <- provider()
+    creds <- merge_in_creds(creds, provider_creds)
```
davidkretch: I may be misunderstanding the implementation, but I don't think we should be merging credentials together.  If we get incomplete credentials from the config, those could be combined inappropriately with credentials from a provider.

davidkretch:
I found [AWS CLI's config/credential file documentation[(https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html), which has both options `source_profile` and `credential_source`, the latter of which allows you to specify Environment or Ec2InstanceMetadata.  I think then maybe what the assume role call should be able to do is call the other providers to get initial credentials.  This might be worth splitting off from the custom config update, which already handles the case [like this](https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/stscreds/) where you use the existing providers to get the role creds from STS.

paws.common/R/client.R
adambanker: Addressed in my last commit

paws.common/R/client.R
adambanker: This is no longer added as a provider but is checked beforehand

paws.common/R/config.R
adambanker: I added a comment. This is because a role session name is required but may not always be specified beforehand. Therefore this creates one for you

paws.common/R/credentials.R
adambanker: No longer does this

paws.common/R/config.R
adambanker: addressed

adambanker:
@davidkretch let me know if you have any ideas to increase code coverage. Otherwise i think this is ready

davidkretch:
@adambanker I'd like to move the config file stuff into another branch, because I don't think the credential providers should all have `aws_profile` since most cannot use it, and I think this means how we call the credential providers needs to change a little. Also I suspect that some of the code can be consolidated since, e.g. we're reading in the config file in six places.

PR #194, opened 2019-09-25
Support custom configuration per service

adambanker:


adambanker:
@davidkretch I got it to create custom services so it does not just overwrite the service in the package name space.

Accordingly
```
ec2east <- paws::ec2(config = list(region = "us-east-1"))
ec2west <- paws::ec2(config = list(region = "us-west-1"))

ec2east$describe_instances()
ec2west$describe_instances()
```
produce different results.

Let me know if you want to scale back the scope of this commit. Thanks

adambanker:
@davidkretch I have addressed your issues

davidkretch:
Unfortunately tooltips no longer appear in RStudio.  Tooltips also don't appear in the previous version (with the `.svc` parameter) and don't appear in a version using R6 classes.

adambanker:
@davidkretch can you give an example?

davidkretch:
If you run the following and hit tab, no tooltips will appear for the operations.

```
acm <- paws::acm()
acm$
```

If you run the following and hit tab, no tooltips will appear for the parameters.

```
acm$add_tags_to_certificate(
```

adambanker:
Are you sure it isn't an issue with something else?
paws_tooltip

davidkretch:
The tooltips taken from the documentation is what is missing.  This is what it should look like:
image

PR #196, opened 2019-10-01
Escape URIs, e.g. S3 query strings with non-alphanumeric characters

davidkretch:
This bug fix allows uploading files to S3 with paths like "a=b/foo".  Previously, characters like "=" were not escaped and this caused the operation to fail.


adambanker:
The code looks good to me and worked when I tested it to upload an object.

PR #199, opened 2019-10-02
Make XML requests skip empty arguments

davidkretch:
Currently, `xml_build_structure` is designed to skip empty arguments, but since it uses `!is.null`, it will process everything that gets a default value like `logical(0)`, which is not null.

This means that if you fail to supply an argument, it will still appear in the request body like `<Foo/>`.

This merge request corrects that by checking for `length(x) > 0`, and also by explicitly converting all scalars of length 0 to NULL.


adambanker:
Good catch. This update looks good to me.

PR #205, opened 2019-10-12
Support custom configuration per service

adambanker:
Add support for custom configuration per service, as in:

```
ec2 <- paws::ec2(config = list(
    credentials = list(
        creds = list(
            access_key_id = "foo",
            secret_access_key = "bar",
            session_token = "baz"
        )
    ),
    endpoint = "https://foo.com",
    region = "us-east-1"
))
```

Changes include:
* Validate (to a limited extent) and set user-provided configuration settings with `set_config`, which adds configuration settings to a service object.
* Get the config from the service object (`ec2` in the above example) with `get_config`.  This is designed to be used by an operation to get configuration that it should use when making requests.
* Credential providers can now accept arguments.  Getting credentials and region from shared credentials/config files will get those corresponding to a profile name given as an argument.  Since different credential providers require different or zero arguments, calling the providers is done with `call_with_args`, which will only call each provider with the appropriate arguments.


adambanker:
@davidkretch this is ready for you to review

davidkretch:
Changes to be made:
* [ ] `new_session` should return the default config.  Custom config should be handled in `client_config`.
* [ ] `new_service`'s `cfgs` param should be optional with default argument null or empty list.  This is so it remains backward compatible with older versions of Paws which won't provide any cfgs argument
* [ ] Check that we have tests for picking up credentials in the right order, e.g. config provided as arguments are used before environment variables.

adambanker:
@davidkretch I have addressed your comments

davidkretch:
Looks good to me.

PR #215, opened 2019-10-17
Replaces double backticks with single backtick in example comments

adambanker:
Replaces double backticks with single backticks in example comments for service functions. This makes it consistent with other R packages.


davidkretch:
I think we want to avoid changing ``` to ``, so I think it should probably be updated to "neither preceded nor followed by a backtick".

davidkretch:
Looks good to me.

PR #222, opened 2019-10-22
Update credentials doc since per-service config is available on CRAN

davidkretch:


adambanker:
We should add that it is only available with paws 0.1.5

Issue #226, opened 2019-12-26
Support Container Providers In Credential Chain


ryanb8:
Paws does not support container_providers ([boto3 class](https://github.com/boto/botocore/blob/13362410ce8aaa5d35b4ffb056c6c5a58d659bb3/botocore/credentials.py#L1792)) in its credential chain ([boto3 credential providers chain](https://github.com/boto/botocore/blob/13362410ce8aaa5d35b4ffb056c6c5a58d659bb3/botocore/credentials.py#L91-L105)).

When using a service like AWS Batch, boto3 will consume and use the permissions tied to the [AWS Batch Job Role (step 9 here)](https://docs.aws.amazon.com/batch/latest/userguide/create-job-definition.html) before using the [ECS Compute Cluster's Instance Role (step 4d here)](https://docs.aws.amazon.com/batch/latest/userguide/create-compute-environment.html). Paws skips over this completely and uses the ecs instance role directly. 

For me this resulted in a several day wild goose chase of why my python docker containers were running with different permissions than my R docker containers.

Can we add this provider to the paws credential chain? Are you open to external PRs? At a minimum I wanted to create this issue to document the difference and help future users out; I'm interested in contributing a fix, though I'm unsure of the time it'll take to do it and if there are extenuating reasons it's not already included. (It looks like you just need to hit ` 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` with that environment variable expanded and get the creds: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html)

davidkretch:
Really sorry about that!  There isn't a reason it isn't already in the credential chain.  Thank you for the references.

We are also completely open to external PRs.  The credential providers are in the paws.common package, which handles building and decoding API calls, specifically in [paws.common/R/credential_providers.R](paws.common/R/credential_providers.R), and the chain itself is defined in [paws.common/R/credentials.R](paws.common/R/credentials.R). A caveat: unfortunately we don't currently have tests for most of the providers. We will also eventually get to adding this new provider, but we're a little busy with the holidays and outside work at the moment.

ryanb8:
Awesome; I'll be making a fix locally for work this week and I'll try to parlay that into a PR here. Thanks for the ultra-prompt response!

ryanb8:
Actually, it looks like y'all are ahead of me! 😅

davidkretch:
[...]

PR #227, opened 2019-12-30
Check the container role for credentials before the instance role

adambanker:
It checks if the associated job role has credentials before checking the instance role. This pull request addresses #226. I tested it by running a batch job with the cran version of paws, as well as with my branch and it only worked on my branch as expected. I also tested out my branch on an ec2 instance with a proper role.

One issue is that running paws on an ec2 instance with a role takes longer because it needs to wait for the job role check to time out first. Let me know if you have any ideas for how to improve this.

### DIFF
--- a/paws.common/R/config.R
+++ b/paws.common/R/config.R
@@ -111,6 +111,27 @@ get_profile_name <- function(profile = "") {
   return(profile)
 }
 
+# Gets the job role credentials by making an http request
+get_job_role_credentials <- function() {
+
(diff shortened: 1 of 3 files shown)


ryanb8:
To avoid the timeout issue, can we avoid making the request to the credentials_uri if the environment variable is empty? I believe R defaults to `''` if the environment variable is not present.

adambanker:
@ryanb8, great call with checking the environment variable! That worked

adambanker:
@davidkretch I made the changes you requested

davidkretch:
I was wrong about how much code they shared, but this works for me and I think this is the right approach.

PR #254, opened 2020-03-06
Sets Git to ignore DS_Store files

adambanker:
On MacOS, .DS_Store files are produced to store the customizations of a folder.
[https://en.wikipedia.org/wiki/.DS_Store](https://en.wikipedia.org/wiki/.DS_Store)

They describe an individual user's environment so they are unnecessary to store in the repo. Furthermore they are added to the `.gitignore` file for other packages such as `dplyr` so there is precedent for this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .Ruserdata
 *.Rcheck
 .vscode
+.DS_Store


davidkretch:
Looks good to me.

PR #255, opened 2020-03-07
Waits to get handlers until run time

adambanker:
I think we should discuss how we want to implement this solution, but this pull request pinpoints the issue where handlers were not changing when we update paws.common.

The issue is that, before, when we were creating a new service object, we were explicitly setting the handlers to be the handlers that existed at package development time. This fix forces the function to use the handlers as they exist at run time by delaying evaluation until then.

I tested this fix by adding a `stop` to `handlers_restxml.R` and reinstalling paws.common. 

```
# Works as normal
restxml_build <- function(request) {
  request <- rest_build(request)
  t <- rest_payload_type(request$params)
  if (t == "structure" || t == "") {
    request <- xml_build_body(request)
  }
  return(request)
}

# Changed it to
# Build the request body for the REST XML protocol.
restxml_build <- function(request) {
  stop("FOR TESTING")
  request <- rest_build(request)
  t <- rest_payload_type(request$params)
  if (t == "structure" || t == "") {
    request <- xml_build_body(request)
  }
  return(request)
}
```

As expected the following command worked without the `stop`, and failed with the `stop`.
```
library(paws)
s3 <- s3()

s3$list_buckets()
```

### DIFF
--- a/make.paws/R/service.R
+++ b/make.paws/R/service.R
@@ -42,10 +42,10 @@ service_file_template <- template(
     target_prefix = ${target_prefix}
   )
 
-  .${service}$handlers <- new_handlers(${protocol}, ${signer})
+  .${service}$handlers <- quote(new_handlers(${protocol}, ${signer}))
 
   .${service}$service <- function(config = list()) {
-    new_service(.${service}$metadata, .${service}$handlers, config)
+    new_service(.${service}$metadata, eval(.${service}$handlers), config)
   }
   `
 )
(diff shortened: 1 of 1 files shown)


davidkretch:
Ah, excellent, that is a simple fix.  It looks good to me.  Do you foresee any issues?  I guess you could potentially call `new_handlers` from within the service function, but that is I think functionally equivalent.

adambanker:
My concerns were mostly stylistic. We haven't really been using quote/eval before. I like moving `new_handlers` because it is more consistent

davidkretch:
Ah, that makes sense.  Looks good to me.

PR #262, opened 2020-03-24
Improve continuous integration

davidkretch:
* Make CI run `R CMD check` on paws.common, make.paws, and paws as CRAN, including running examples marked `donttest`.
* Add a `make check` recipe to the makefile, which runs `R CMD check` on paws.common and all CRAN packages.
* Change `dontest` examples in paws.common to `dontrun` to avoid errors in `R CMD check`.
* Delete AppVeyor.


davidkretch:
@adambanker What do you think?  I think `dontrun` is the way to go here.  It seems to have worked for paws.common.

adambanker:
That seems to be more in the spirit with what the intention for those tests are, so I am for doing it that way.

davidkretch:
Do you think I should fix all the Paws examples in this PR or a separate PR?  This PR will fail CI without updating Paws, but updating Paws will be a lot of small changes.

adambanker:
We can do it in a separate PR so it is more clear what this PR accomplishes

davidkretch:
Ok, cool.  Let me know how it looks.

adambanker:
@davidkretch why did you delete appveyor?

davidkretch:
We've had a lot of problems with it in the past, and I figured we'd eventually switch to GitHub Actions.  I can add it back.

adambanker:
I think we should add it back because we won't be testing on windows without it. But yes moving to github actions would be even better

davidkretch:
Cool, I restored AppVeyor.  Eventually we'll move to GitHub Actions but it's more than I want to do atm.

PR #263, opened 2020-03-28
Use `dontrun` instead of `donttest`; escape @, {, and } in documentation

davidkretch:
* Use `dontrun` in examples instead of `donttest` to avoid errors in `R CMD check`.  Under `donttest`, CRAN still runs examples and throws errors when there are examples that fail.
* Escape `@` in documentation (as `@@`), since it is a special character for Roxygen. Source: "Because @ has a special meaning in roxygen, you need to write @@ if you want to add a literal @ to the documentation" in http://r-pkgs.had.co.nz/man.html#roxygen-comments.
* Escape `{` and `}` in strings included in examples, since these can cause errors in LaTeX.


davidkretch:
@adambanker This also passes `R CMD check` using `make check`.

adambanker:
Looks good to me!

Issue #271, opened 2020-04-06
Error in interface[[1]] : subscript out of bounds on empty response


jrdi:
I'm running a `put_metric_data` and getting the following `Error in interface[[1]] : subscript out of bounds`. Since CloudWatch is showing the data I have been debugging the response parsing.

I've found that in `xml_unmarshal` the first node is retrieved despite the name of the field and it happens to be the metadata when the API returns an empty response.

```r
# Unmarshal `data` provided as a list into the shape in `interface`.
xml_unmarshal <- function(data, interface, result_name = NULL) {
  root <- data[[1]]
  if (!is.null(result_name) && result_name %in% names(root)) {
    root <- root[[result_name]]
  }
  result <- xml_parse(root, interface)
  return(result)
}
```

![image](https://user-images.githubusercontent.com/67608/78565850-bd63ae80-781e-11ea-884c-e7c3cf350978.png)

Then when `xml_parse` is called (and then `xml_parse_list`) the node contains data but the interface object is requiring an empty list. Here you have the actual line that is crashing.

```r
xml_parse_list <- function(node, interface) {
  if (length(node) == 0) return(list())
  names(node) <- NULL
  result <- lapply(node, function(x) xml_parse(x, interface[[1]]))       <-----
  if (type(interface[[1]]) == "scalar") {
    result <- unlist(result)
  }
  return(result)
}
```

Shouldn't `xml_unmarshal` skip metadata node before extracting the first node by default?

davidkretch:
Sorry about that but thank you for the bug report!  We'll try to fix this very soon.  Do you also have an example call to `put_metric_data` that we could try?

jrdi:
Hey @davidkretch, thank you for your reply. Sure, here you have the exact snippet I'm using. Let me know if you need something else to reproduce it or if you prefer I can try to submit a PR with the patch.

```r
cloudwatch <- paws::cloudwatch()
cloudwatch$put_metric_data(Namespace = "Test.Namespace", list(
  list(
    MetricName = "Metric1",
    Value = 10,
    Unit = "Count"
  ),
  list(
    MetricName = "Metric2",
    Value = 20,
    Unit = "Count"
  )
))
```

davidkretch:
Thanks!  It looks like we should not be calling the unmarshal handler at all in this case since the interface (i.e. the API response shape) is empty.  Sorry -- I put together a small fix.  I'll update this issue when a fix is available in master.

davidkretch:
[...]

PR #273, opened 2020-04-09
Do not run unmarshal handlers if the output shape is empty

davidkretch:
If the output shape is empty, i.e. it has no elements, then don't try to unmarshal the API response.

Addresses #271.

## Example

For example, in CloudWatch `PutMetricAlarm`, its output shape is empty -- it has no data elements -- but the API response contains some metadata. When its unmarshal handler gets called, the handler fails. With this update, it skips calling the unmarshal handler since there is nothing to unmarshal into.

### `PutMetricAlarm`'s output shape:

```
.cloudwatch$put_metric_alarm_output <- function(...) {
  list()
}
```

### `PutMetricAlarm` API response:

```
<PutMetricDataResponse xmlns=\"http://monitoring.amazonaws.com/doc/2010-08-01/\">
  <ResponseMetadata>
    <RequestId>317752ff-66e9-4f98-b643-661a35fff360</RequestId>
  </ResponseMetadata>
</PutMetricDataResponse>
```

### DIFF
--- a/paws.common/R/handlers.R
+++ b/paws.common/R/handlers.R
@@ -154,6 +154,7 @@ validate_response <- function(request) {
 }
 
 unmarshal <- function(request) {
+  if (length(request$data) == 0) return(request)
   request <- run(request, unmarshal)
   return(request)
 }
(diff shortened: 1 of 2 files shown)


adambanker:
Looks good

Issue #272, opened 2020-04-07
s3 select_object_content: Error: Root nodes must be of length 1


DyfanJones:
Hi all,

I am trying to select a subset of data from an s3 object, in this example `iris` dataset.  Unfortunately I am coming across the following error when using `select_object_content`:

```
s3 <- paws::s3()

obj <- s3$select_object_content(Bucket = "mybucket",
                                Key = "iris.csv.gz",
                                Expression = "select * from s3object limit 11",
                                ExpressionType = "SQL",
                                InputSerialization = list('CompressionType' = "GZIP",
                                                          'CSV' = list(
                                                            'RecordDelimiter'= '\n',
                                                            'FieldDelimiter'= ',')),
                                OutputSerialization = list('CSV'= list('RecordDelimiter'= '\n',
                                                                       'FieldDelimiter'= ',')))
```

Error: `Error: Root nodes must be of length 1`

----------

Session information:
```
devtools::session_info()
─ Session info ───────
 setting  value                       
 version  R version 3.6.1 (2019-07-05)
 os       macOS High Sierra 10.13.6   
 system   x86_64, darwin15.6.0        
 ui       RStudio                     
 language (EN)                        
 collate  en_GB.UTF-8                 
 ctype    en_GB.UTF-8                 
 tz       Europe/London               
 date     2020-04-07                  

─ Packages ──────────
 package      * version date       lib source        
 assertthat     0.2.1   2019-03-21 [2] CRAN (R 3.6.1)
 backports      1.1.5   2019-10-02 [1] CRAN (R 3.6.0)
 callr          3.4.1   2020-01-24 [2] CRAN (R 3.6.0)
 cli            2.0.1   2020-01-08 [2] CRAN (R 3.6.0)
 crayon         1.3.4   2017-09-16 [2] CRAN (R 3.6.1)
 curl           4.3     2019-12-02 [1] CRAN (R 3.6.0)
 desc           1.2.0   2018-05-01 [2] CRAN (R 3.6.1)
 devtools       2.2.1   2019-09-24 [2] CRAN (R 3.6.0)
 digest         0.6.25  2020-02-23 [1] CRAN (R 3.6.0)
 ellipsis       0.3.0   2019-09-20 [1] CRAN (R 3.6.0)
 fansi          0.4.1   2020-01-08 [2] CRAN (R 3.6.0)
 fs             1.3.1   2019-05-06 [2] CRAN (R 3.6.1)
[...]

PR #275, opened 2020-04-10
Add metadata to input/output shapes

davidkretch:
* Add `locationName` and other metadata to input and output shapes. For example, in the S3 `SelectObjectContent` operation, the request body contains a number of elements that should be within a `<SelectObjectContentRequest>` XML root node. Without the `locationName`, Paws won't create this root node and will then fail.

Addresses the bug in #272.  However, once this is fixed, we still need to support [`SelectObjectContent`'s chunked transfer encoding](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html).


adambanker:
Looks good to me

Issue #280, opened 2020-04-17
Adding Support for Temporary Credentials (Session Token) in `credentials_file_provider()`


CurryEleison:
Paws supports temporary AWS credentials with session tokens _except_ in the case where credentials are provided in a credentials file. Unfortunately for me, that's where my session token is stored :)

I think it would be an easy improvement. In the `credentials_file_provider()` method of `credential_providers.R` we just need to read out the session token with 
```
session_token <- credentials[[aws_profile]]$aws_session_token
```
and pass the result to `creds()`.

For context, the case where session token is stored in the environment was improved in #185, and there's an excellent workaround in #210 , so this is clearly not a high priority -- it would just make authentication a little more consistent.

adambanker:
Thank you for submitting the issue! This looks like an oversight on our part, but your proposed fix looks like it should solve the issue. I can implement the fix and put it in the next Paws update unless you would prefer to submit your own pull request.

CurryEleison:
Thanks! It's probably going to be faster/better if you implement it :)

adambanker:
Hello, the fix is now available on GitHub.  You can get the fixed version by running `devtools::install_github("paws-r/paws/paws.common")`. It will be available on CRAN in the next release, but you may have to wait a few weeks for that since there is a limit to how often they accept updates.

CurryEleison:
That seems to work perfectly! I did some manual testing in various scenarios and everything worked as expected. Thanks!

davidkretch:
Hello, this bug fix is now in the version on CRAN.  Let us know if we can close this issue.

Thank you!

CurryEleison:
Just did some manual tests on this from CRAN and it works perfectly 👍

PR #282, opened 2020-04-18
Reads in session token from credentials file

adambanker:
Other SDKs support reading in a session token from the credentials file. This addresses #280 and enables that capability.


davidkretch:
Looks good to me.  I'm ok with a drop in coverage too.

Issue #276, opened 2020-04-11
Support event streams / chunked transfer encoding


davidkretch:
See https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html.

davidkretch:
Hi @DyfanJones, just wanted to let you know I've made some progress on this and hope to have it working in the next couple weeks.  Hope all is well over there.  🙂

DyfanJones:
@davidkretch this is great news! It is really exciting to see the quick development of R SDK into AWS. Thanks for all the hard work you and @adambanker  are putting into it.

I hope everything is ok with you and your family too :)

davidkretch:
Hi @DyfanJones,

I have `select_object_content` working on branch `feature/streams`.  If you want to try it, you can get it with `devtools::install_github('paws-r/paws/paws.common@feature/streams')`.  The two questions I have are:

1. Is performance acceptable?
2. Should it also decode the CSV/JSON response into data frames?

DyfanJones:
Hi @davidkretch,

I am coming across the following error when I am testing your new branch:

```
s3 <- paws::s3()

obj <- s3$select_object_content(Bucket = "my-bucket",
                                Key = "iris.csv.gz",
                                Expression = "select * from s3object limit 10",
                                ExpressionType = "SQL",
                                InputSerialization = list('CSV' = list("RecordDelimiter"= '\n',
                                                                       "FieldDelimiter"= ','),
                                                          'CompressionType' = "GZIP"),
                                OutputSerialization = list('CSV'= list("RecordDelimiter"= '\n',
                                                                       "FieldDelimiter"= ',')))
```
```
Error: MissingRequiredParameter (HTTP 400). RecordDelimiter is required. Please check the service documentation and try again.
```

To answer your questions:
1) If the performance is roughly same as boto3 then it should be fine :)
2) I don't think decoding should be necessary as boto3 doesn't do it and leaves it to the user. Maybe some documentation should be enough?

davidkretch:
Hello @DyfanJones,

This functionality is now in the version of `paws.common` on CRAN.

With respect to the issue with `RecordDelimiter` `\n`, that is a separate bug which I created a new issue for. The default record delimited is `\n` so you should be able to leave it out for now.

I put an example in [examples/s3_select_object_content.R](https://github.com/paws-r/paws/blob/master/examples/s3_select_object_content.R). We still have to figure out a better way to integrate these sorts of examples with the rest of the package.

[...]

PR #295, opened 2020-06-21
Add support for event streams and S3 SelectObjectContent

davidkretch:
Decode event streams as used in the S3 SelectObjectContent operation, which allows you to query objects in S3 using SQL and return them as delimited text e.g. CSV.


davidkretch:
Addresses #276.

PR #305, opened 2020-07-18
Add Dyfan's SageMaker example

davidkretch:


davidkretch:
Hello @DyfanJones, what platform are you using? I am on Windows using R 4.0.0 and keep getting the error `label must be in [0,1] for logistic regression`. I get this even if I use `fwrite` argument `eol="\n"`, change the dataset to `iris`, use Python, etc. Did you run into anything like this?

DyfanJones:
Hi @davidkretch I ran the initial example through AWS Sagemaker. My gut feels is there is an issue with the target variable however. I am happy to double check this.

DyfanJones:
Hi @davidkretch  I have managed to get an iris data set example up and running [sagemaker-xgboost-example-02-paws](https://github.com/DyfanJones/r-sagemaker-examples/blob/master/R/sagemaker-xgboost-example-02-paws.R). I have also reduced the number of dependencies so that it is just paws.

What I believe was happening is that the xgboost "binary:logistic" model was looking for a binary target and got [0,1,2] target instead. In the new example I have gone for xgboost "multi:softprob" model which can cope with none binary targets. The new model has been developed from [XGBoost Multinomial Classification Iris Example in R](https://rpubs.com/dalekube/XGBoost-Iris-Classification-Example-in-R) and should help users relate Sagemaker xgboost tothe xgboost package.

Let me know if it works for you :) happy to walk through if need be :)

PR #310, opened 2020-07-24
Add GitHub URL

lorenzwalthert:
This helps when one wants to go to GitHub from the CRAN package page. There is currently no back-linking: 
image


davidkretch:
Hey, thank you for the PR! Also, sorry for the delayed response :-(

The URL on CRAN comes from [this file](https://github.com/paws-r/paws/blob/master/cran/paws/DESCRIPTION), which is automatically generated by the [`write_description_category`](https://github.com/paws-r/paws/blob/master/make.paws/R/package.R#L15) function. I'll update that in this PR if that's ok with you.

lorenzwalthert:
Sure.

PR #315, opened 2020-08-07
example of how to access Federated API using SAML 2.0 and AD FS 

DyfanJones:
Came across this example in how to access Federation API using SAML2.0 and AD FS https://aws.amazon.com/blogs/security/how-to-implement-federated-api-and-cli-access-using-saml-2-0-and-ad-fs/ .  Created a simple wrapper function with a paws::sts and paws::s3 example.


davidkretch:
Very cool! Thank you.

Issue #329, opened 2020-09-19
Error in stop(aws_error(request$error)) : bad error message


BillWeld:
The error object contains useful information in its attributes but not in its field values.

Love the library btw.  Super useful!

[tmp.txt](https://github.com/paws-r/paws/files/5250017/tmp.txt)

davidkretch:
Yeah, haha, that is a bad error message. Either I or someone else will try to get to this soon. Thank you for the heads up!

davidkretch:
Hello again, is there an example from another package I can look at where try errors have info in fields? I have been trying and not been able to get it to work -- it seems like `try` puts them in the `condition` attribute.

davidkretch:
Disregard my comment. I understand now.

davidkretch:
Hello again. This should be fixed with the next version of `paws.common` on CRAN in about 2 weeks, due to CRAN rules about how often we can update. Thank you again for the bug report!

BillWeld:
Thanks for a super useful library!

davidkretch:
Sorry for the delay. This is now on CRAN. Let us know if you have any issues! 

Thank you!

PR #336, opened 2020-11-02
Add status code and error response to condition object

davidkretch:
* Add the error response and HTTP status code to the condition object, in addition to having it be in the condition object's attributes. This is in response to a request in issue #329.
* Make errors be of class `paws_error` in addition to their other classes, e.g. `http_400`, `error`, `condition`.


adambanker:
Looks good!

PR #346, opened 2020-12-10
Fix warning: "partial match of 'host' to 'hostname'"

jcheng5:
This occurs when options(warnPartialMatchDollar=TRUE) and you do almost any
operation.

Example:

```r
options(warnPartialMatchDollar = TRUE)
library(paws)
svc <- s3()
svc$list_buckets()
```

```
Warning messages:
1: partial match of 'host' to 'hostname' 
2: partial match of 'host' to 'hostname' 
3: partial match of 'host' to 'hostname' 
```

### DIFF
--- a/paws.common/R/url.R
+++ b/paws.common/R/url.R
@@ -19,8 +19,8 @@ Url <- struct(
 parse_url <- function(url) {
   p <- httr::parse_url(url)
   if (is.null(p$scheme)) p$scheme <- ""
-  if (is.null(p$host)) p$host <- ""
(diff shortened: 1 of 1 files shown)


adambanker:
Thank you for submitting this PR!

@davidkretch created [structs](https://github.com/paws-r/paws/blob/main/paws.common/R/struct.R) for this package to catch bugs caused by accessing list items by the wrong name. It looks like it did not catch that here because this list came externally from `httr`, and we were grabbing the hostname so that we could put it into the Url struct.

This fix looks good to me. Thanks!

PR #347, opened 2020-12-12
Fix `make deps`

jcheng5:
I saw two problems with `make deps` off a fresh git clone on macOS:

1. Always emits the message about installing pandoc, even though it's on my path.
   I don't actually know what was wrong with the old incantation, which worked
   correctly for me at the bash shell but not within the Makefile; anyway, the
   new one works reliably for me.

2. Submodule is not initialized; this is because vendor/aws-sdk-js/ exists on a
   fresh clone, it's just empty.


davidkretch:
Looks good to me. Thank you!

PR #348, opened 2020-12-13
make_sdk() speedup

jcheng5:
A great deal of time in `make_sdk()` is spent in two classes of operations:

1. `url_ok()`
2. Converting HTML to text or markdown using pandoc

These operations are pretty slow, not dependent on side effects, and their
implementations are pretty stable at this point (maybe?), so they're good
candidates for persistent caching.

This commit uses the cachem package to manage a ./cache subdirectory for
uses of `url_ok()` and `html_to()` (update: and `convert()`).

On my MBP15 from 2016, I got the following times for `make_sdk()`:

- Cache disabled:         62m17.732s
- With cache, first run:  43m11.440s
- With cache, second run: 10m0.548s (update: now down to 8m34.990s)

(The "first run" starts with the cache empty, but there are enough
duplicated `html_to()` calls that it still gets some cache hits.)

If the implementations of `url_ok()` or `html_to()` change, you'll need to
throw out the cached values by simply deleting the ./cache directory;
you can use `make clean` for this. I gave the caching logic a "spot
checking" feature to try to proactively detect this condition; by default
1% of cache hits are compared to the value that's yielded by running
the operation, and if the values are not identical, an error is thrown.

(I also added default arguments for `make_sdk()`; I didn't see a downside to
doing this, and it made iterative profiling a bit less annoying.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 *.Rcheck
 .vscode
 .DS_Store
+cache
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 IN_DIR := ./vendor/aws-sdk-js
 OUT_DIR := ./paws
 CRAN_DIR := ./cran
+CACHE_DIR := ./cache
 
 # Make R use the user's package library by setting the R user home path (R_USER)
 # to the folder containing their package library. On Windows, it is in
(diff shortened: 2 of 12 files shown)


davidkretch:
This is great. Thank you!

PR #350, opened 2020-12-15
Disable byte compilation of paws

jcheng5:
On my MBP15, disabling byte compilation drops install time
from 5m58s to 1m34s. It seems unlikely this will affect
runtime performance, given the kind of R code that paws
contains.

Note that paws.common and cran packages are not affected.


davidkretch:
Looks good to me. Thanks!

PR #351, opened 2020-12-17
Remove warnings during make_sdk

jcheng5:
* With options(warnPartialMatchDollar=TRUE), make_sdk was warning about
  "member" versus "members"
* Also warning about file("") (empty string), due to system.file()
  results being unconditionally readLines()'d
  (file("") only supports open = "w+" and open = "w+b": using the former)


davidkretch:
Looks good to me.

PR #352, opened 2020-12-17
Use short timeout when requesting instance/container metadata

jcheng5:
These requests took several long seconds to fail when called from
non-AWS. The default in aws-sdk-js-v3, boto3, and aws-cli is 1
second, no retries.

Ref:
https://github.com/boto/botocore/blob/646c61a7065933e75bab545b785e6098bc94c081/botocore/utils.py#L273
https://github.com/aws/aws-sdk-js-v3/blob/47770fa493c3405f193069cd18319882529ff484/packages/credential-provider-imds/src/remoteProvider/RemoteProviderInit.ts#L1


davidkretch:
Looks good to me.

PR #353, opened 2020-12-19
Remove most stray backslashes in docs

jcheng5:
Fixes stray '\' characters that appear everywhere in the docs. This
is due to pandoc's overly-aggressive markdown escaping in `markdown`
output mode. With `commonmark` mode pandoc does this an order of
magnitude less, but there are still some left. Grepping for
`'\\\\[^a-zA-Z]'` in the ./paws/man directory shows them.

The first two commits contain the actual code changes, which are extremely tiny. The thousands of files changed are all in the third commit, which is just the results of `make build`.

**Update:** In order to get tests to pass I needed to pin the version of pandoc to something newer. It might be a good idea to be running pandoc v2.11.x locally if you're not already (daily builds of RStudio IDE include v2.11.x).


davidkretch:
This looks good to me, but it also revealed some Unicode encoding issues on Windows. I'll have fixes on another branch [`bug/windows-utf8-encoding`](https://github.com/paws-r/paws/commit/9d1c5ef200e417d773c3ea270f488ca10b5b2afc).

PR #355, opened 2020-12-27
Fix make.paws' handling of Unicode on Windows

davidkretch:
* Explicitly specify UTF-8 encoding when reading in text converted with Pandoc. With the change to Commonmark from Markdown, some text with Unicode would get read in incorrectly on Windows.
* When making LaTeX tables for the documentation in `tabular`, do *not* call `format`. Since all data is already character strings, `format` is unnecessary and on Windows it re-encodes from UTF-8 to another encoding, and this results in incorrect output.


davidkretch:
Hello @jcheng5, this PR fixes some Unicode encoding issues that appeared in Windows. If you have time, would you be willing to check that it works for you as well?

jcheng5:
make.paws unit tests pass on macOS.

The only differences I see in the generated docs are:

1. In the `\tabular`, each row used to have a space before `\cr`, now they don't.
image

2. Some links work now where they didn't before. 
image

jcheng5 (approved):

PR #363, opened 2021-01-16
Allows you to supply a file path as an argument to s3 put object

adambanker:
Previously when using `s3$put_object` to upload a file to s3, you had to convert the file to a raw object first. You would then supply that raw object to the Body parameter of the put_object function.

With this update, if the user enters a file path for the body, the function will automatically convert it to a raw object for you.
```R
## New syntax
s3 <- s3()
bucket_name <- "my-example-bucket"
file_name <- "example.rds"
s3$put_object(
  Body = file_name,
  Bucket = bucket_name,
  Key = file_name
)
```

### DIFF
--- a/paws.common/R/custom_s3.R
+++ b/paws.common/R/custom_s3.R
@@ -4,6 +4,28 @@ NULL
 
 ################################################################################
 
+convert_file_to_raw <- function(request) {
+  operation_name <- request$operation$name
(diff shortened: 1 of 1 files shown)


davidkretch:
Looks good to me.

davidkretch (approved):

PR #367, opened 2021-01-20
Updates s3 example to use the new simplified syntax

adambanker:
Update the s3 example to show how to upload a file by just passing in the file path.


davidkretch (approved):
Looks good to me.

PR #369, opened 2021-01-23
Cache credentials

davidkretch:
* Cache credentials in the `.internal$config$credentials` object by making it an environment. Whenever an operation is run, the SDK will add credentials to the credentials environment when needed. These credentials will be cached and available in the service object, e.g. the object created by `paws::s3()`.
* Check for credential expiration by adding an expiration property to the `Creds` object.

TODO:
* [x] Test on AWS
    * ~~EC2 role~~
    * ~~container role~~
    * ~~assume role~~
    * ~~credential process~~
* [x] Add time windows to expiration timestamp?
* [x] Add unit tests

### DIFF
--- a/paws.common/tests/testthat/test_credentials.R
+++ b/paws.common/tests/testthat/test_credentials.R
@@ -131,7 +131,7 @@ test_that("credentials refresh when expired", {
 
   expect_equal(expiration1, expiration2)
 
-  Sys.sleep(5) # credentials expire after 5 seconds
+  Sys.sleep(5) # these credentials are set to expire after 5 seconds
 
   f3 <- svc$get_credentials() # refresh the credentials
   expiration3 <- f3$credentials$creds$expiration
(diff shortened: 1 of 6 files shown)


paws.common/tests/testthat/test_credentials.R
```diff
@@ -14,24 +14,130 @@ test_that("credentials not provided", {
...
+})
+
+#-------------------------------------------------------------------------------
+
+# Example operation that gets credentials (if unavailable or expired) and
+# returns the credentials.
+f_get_credentials <- function() {
+  cfg <- get_config()
+  if (!is_credentials_provided(cfg$credentials$creds, window = 0)) {
+    foo <- get_credentials(cfg$credentials)
```
adambanker, line 73: This test works but I am a little confused about what is happening since `foo` isn't being used. Is it because `get_credentials` sets the value and `cfg$credentials` is just a reference to that environment so it will still be pointed at the correct value?

adambanker (approved):
This update looks good to merge! I had one question since I haven't done much with environments before, but overall this is a great solution!

paws.common/tests/testthat/test_credentials.R
davidkretch: That's exactly right. I'll add comments that it sets the credentials as a side effect, and get rid of the assignment to `foo`.

Issue #253, opened 2020-03-02
Support for Assume Role Provider


davidski:
It would be wonderful if `paws` could support the Assume Role Provider functionality present in boto3/botocore and documented at https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html. Specifically, I'm looking to find support for `paws` to read and honor the `source_profile`, `mfa_serial`,  and `credential_process` parameters. With these three, paws could be used to transparently assume a role where credentials are not stored on disk in plaintext, but instead are retrieved via secure storage with a  program such as [aws-vault](https://github.com/99designs/aws-vault), assuming a role with temporary credentials.

This is a complicated request, but is currently a missing feature in all of the R AWS packages and would be a tremendous add.

davidkretch:
Thanks, yes, we'll try to work on that soon.  We started on part of it a while back actually.  I can't say when it'll be done though.

dpastoor:
For those of you that stumble onto this, this was the workaround I came up with as a workaround:

```
get_creds_assumed_profile <- function(.x) {
  .profile_details <- ini::read.ini("~/.aws/credentials")[[.x]]
  if (is.null(.profile_details)) {
    stop(sprintf("no profile with the name `%s` found in ~/.aws/credentials file", .x))
  }
  profile <- .profile_details$source_profile
  # if no source_profile lets look for a credential_source, which will probably
  # be something like Ec2InstanceMetadata
  if (is.null(profile)) {
    profile <- .profile_details$credential_source
  }
  if (is.null(profile)) {
    stop("could not find either a credential_source or source_profile, please make sure one is set!")
  }
  sts <-paws::sts(config = list(credentials = 
                                  list(profile = profile), 
                                region = "us-east-1"))
  .pc <- sts$assume_role(.profile_details$role_arn, "assumed-role")
  return(list(
    access_key_id = .pc$Credentials$AccessKeyId,
    secret_access_key = .pc$Credentials$SecretAccessKey,
    session_token = .pc$Credentials$SessionToken
  ))
}
```

```
creds <-  get_creds_assumed_profile("techopsdev")
svc <- sts(config = list(credentials = list(creds = creds)))
svc$get_caller_identity() 
s3_svc <- s3(config = list(credentials = list(creds = creds)))
s3_svc$list_buckets()
```

From a quick look at the PR, this seems to be the direction the PR is going, just only had ~ 1 hr to come up with something for my team to mess with.

[...]

PR #373, opened 2021-01-29
Support multifactor auth

davidkretch:
Support multifactor authentication, using the `mfa_serial` shared configuration file item.

Addresses #253

To do:
* [x] Add comments
* [x] Improve prompt?
* [x] Test on AWS
    + ~~EC2 IAM role~~
    + ~~config file credential process~~
    + ~~config file assume role: source profile~~
    + ~~config file assume role: credential source EC2 instance metadata~~
    + ~~config file assume role: source profile + MFA~~


adambanker (approved):
Looks good! I think this will be a really useful feature

PR #374, opened 2021-01-30
Link to the Paws help pages for all API operations

davidkretch:
* Wherever the documentation links to an API operation's help, e.g. `ListBuckets`, convert it to a link to the corresponding Paws help page, e.g. for `list_buckets`.

  Do this in `convert` and in the `clean_html_xxx` functions by checking whether each link or code snippet matches the name of one of the API operations. If so, replace the name and point to the corresponding Paws help page. To recognize API operations, pass a list of operations to the `links` parameter.

  This is done in two parts: 1) Replace the link and link text, prior to Markdown conversion, and 2) After Markdown conversion, search for ordinary links that look like they *should* be Roxygen help links, e.g. `[operation_name](svc_operation_name)`, and replace them with valid Roxygen help links, e.g. `[operation_name][svc_operation_name]`. (first commit)

* Due to the internal linking above, there is now identical documentation from two different services (`waf` and `wafregional`) that produces *different* outputs when converted into Markdown. The caching used to speed this up this conversion now gets a `service` key to avoid collisions and spot check failures in these cases.

* Regenerate Paws. (second commit)


davidkretch:
Hey Adam, if you have the time, could you review this? To see what I was going for, I recommend looking at the new tests in the first commit.

make.paws/R/docs.R
```diff
@@ -199,81 +199,138 @@ comment <- function(s, char = "#") {
...
+
+  # Replace API operation names with links to corresponding R function names.
+  internal_link <- links[[trimws(text)]]
+  if (!is.null(internal_link)) {
+
+    # Don't add a link if this node is already within a link.
+    parent <- xml2::xml_parent(node)
+    if (xml2::xml_name(parent) == "a") {
+      xml2::xml_text(node) <- internal_link$r_name
+      return()
```
adambanker, line 285: To be consistent, we usually do `return(NULL)` instead of just `return()`

adambanker (commented):
Looks good! I noticed some typos/style things, but other than that it worked for me.

adambanker (approved):

PR #378, opened 2021-02-07
Remove extra escapes, fix overly long titles

davidkretch:
* Remove most extra escapes. Now the only escaping done is for the following cases:
    * All `{` and `}` in plain text to avoid LaTeX errors.
    * Unmatched `{` and `}` in code fragments to avoid Roxygen errors.
    * Unmatched `"`, `'`, or ` in code fragments to avoid Roxygen errors.
* Explicitly remove extra escapes added by Pandoc for `_` and for `\` before `{` and `}`. Explicitly remove extra escapes in generated example code.
* Clean up the HTML-to-Markdown pipeline, which had a lot of extra complexity that is no longer needed since moving to Commonmark.
    * In `convert`, treat all provided documentation as HTML. Previously it would check for first character `<` and if not found, would treat as plain text. However, some documentation had extra whitespace before the HTML began. Now it trims whitespace, and if the text does not begin with an HTML tag, it wraps it in a `<body>` element.
* Fix long help page titles caused by documentation whose first paragraph doesn't end with a period. Previously it would include the next paragraph; now it does not.
* Do not run Roxygen for the original `paws` folder; only for the packages in the `cran` folder. This was not needed and avoiding it saves build time.

### DIFF
--- a/make.paws/R/make_sdk.R
+++ b/make.paws/R/make_sdk.R
@@ -16,7 +16,6 @@ make_sdk <- function(in_dir = "./vendor/aws-sdk-js", out_dir = "./paws",
       cat(paste0(api, "\n"))
       write_sdk_for_api(api, in_dir, out_dir)
     }
-    write_documentation(out_dir)
     return(invisible(TRUE))
   })
 }
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,8 @@ export R_USER
 
 help:
 	@echo "  build              build the AWS SDK package"
-	@echo "  install            build and install the AWS SDK package"
+	@echo "  rebuild            clear the build cache and build"
+	@echo "  install            build and install the AWS SDK packages"
 	@echo "  common             build and install common functions"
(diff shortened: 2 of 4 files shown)


make.paws/R/make_sdk.R
```diff
@@ -16,7 +16,6 @@ make_sdk <- function(in_dir = "./vendor/aws-sdk-js", out_dir = "./paws",
       cat(paste0(api, "\n"))
       write_sdk_for_api(api, in_dir, out_dir)
     }
-    write_documentation(out_dir)
```
adambanker, line 19: One issue with only removing this line is that it appears that the DESCRIPTION file does not get rewritten in the paws folder. I noticed this because I tried to run make_sdk on a subset of aws services and kept getting an error because it was expecting files in the collate section that were not there.

davidkretch: Updated the Makefile `install` recipe.

adambanker (approved):

PR #384, opened 2021-02-14
Support more config and credential options

davidkretch:
* Support `AWS_CONFIG_FILE`, `AWS_SHARED_CREDENTIALS_FILE`, `AWS_CREDENTIAL_EXPIRATION`, and `AWS_EC2_METADATA_DISABLED`.

* Update the credentials documentation.

* Add `get_env` function to get environment variables from either R or the OS. Change nearly all instances of `Sys.getenv` to `get_env`. Combine `r_env_provider` and `os_env_provider` credential providers. Delete `check_r_env_region` and `check_os_env_region`.

* Move credential related functions `get_container_credentials` and `get_iam_role` to `credential_providers.R`

### DIFF
--- a/docs/credentials.md
+++ b/docs/credentials.md
@@ -4,17 +4,18 @@
   - [Set credentials for all services with environment variables](#set-credentials-for-all-services-with-environment-variables)
   - [Get credentials from the AWS shared credentials file](#get-credentials-from-the-aws-shared-credentials-file)
   - [Set credentials for an individual service](#set-credentials-for-an-individual-service)
-  - [Get credentials from an EC2 or container IAM role](#get-credentials-from-an-ec2-or-container-iam-role)
+  - [Get credentials from an EC2 instance or container role](#get-credentials-from-an-ec2-instance-or-container-role)
   - [Get credentials from a command line process](#get-credentials-from-a-command-line-process)
-  - [Assume an IAM role with credentials from the environment or an IAM role](#assume-an-iam-role-with-credentials-from-the-environment-or-an-iam-role)
-  - [Assume an IAM role with credentials from another profile](#assume-an-iam-role-with-credentials-from-another-profile)
-  - [Use multifactor authentication](#use-multifactor-authentication)
+  - [Assume a role with credentials from the environment or an instance/container role](#assume-a-role-with-credentials-from-the-environment-or-an-instancecontainer-role)
+  - [Assume a role with credentials from another profile](#assume-a-role-with-credentials-from-another-profile)
+  - [Use multifactor authentication when assuming a role](#use-multifactor-authentication-when-assuming-a-role)
 - [Set region](#set-region)
   - [Set region for all services with an environment variable](#set-region-for-all-services-with-an-environment-variable)
   - [Get region from the AWS config file](#get-region-from-the-aws-config-file)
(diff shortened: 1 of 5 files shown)


paws.common/R/config.R
```diff
@@ -95,61 +99,69 @@ get_aws_path <- function() {
...
-    issue(metadata_request)
-  }, error = function (e){
-    NULL
-  })
+# Gets the instance metadata by making an http request.
+get_instance_metadata <- function(query_path = "") {
 
-  if (is.null(metadata_response) || metadata_response$status_code != 200) {
+  # Do not get metadata when the disabled setting is on.
+  if (trimws(get_env("AWS_EC2_METADATA_DISABLED")) %in% c("true", "1")) {
```
davidkretch, line 161: Enforce lower case.

docs/credentials.md
```diff
@@ -1,250 +1,508 @@
...
+# Set credentials
 
-If you are running the package on an instance with an appropriate IAM
-role, Paws will use it automatically and you don’t need to do anything
-extra.
+In order to use AWS APIs, you must provide your credentials. Paws will look for
+your credentials in the following order.
 
-### Setting Credentials with Custom Service Config
+1.  In settings provided to an individual service.
```
adambanker, line 31: It is a little confusing that we list the order here:
1) By service
2) By environment variables
3) In files
4) IAM role

but then then the information isn't displayed in that order and instead as:

1) Env variables
2) File
3) By service
4) IAM role
5) Command line
6) Assume Role
7) Multifactor

We should:
* make that ordering consistent
* make these links to the corresponding section
* add the other options to this list

docs/credentials.md
```diff
@@ -1,250 +1,508 @@
...
+
+---
+
+
+# Set region
+
+To use AWS, you must also set your AWS region. Paws will look for the region in
+the following places, in order:
+
+1.  In settings provided to an individual service.
```
adambanker, line 260: Similar issue with ordering as with credentials. These should also be links

adambanker (commented):
The updates look good, but I have some comments about the documentation

docs/credentials.md
davidkretch: I think then we should take that 4 point list out, because it appears again more appropriately in the Reference section and its aim is to explain in what order different sources are checked. Whereas the goal of the first half of the document is to explain how to address a specific use case (how to).

adambanker: I think that is fine. I think we should add a header to that later section, though, so it is easier to find.

docs/credentials.md
```diff
@@ -1,250 +1,525 @@
...
-    
-    [profile other_profile]
-    region=region-2
 
-## AWS Profile
+# Reference
 
-If you do not set your AWS profile, the package will use the `default`
-profile.
+Credentials and options can be set in the following locations. Paws will check
```
adambanker, line 364: We should add a header for this section with a name like
```
## Credential priority
```

docs/credentials.md
davidkretch: Moved to reference.

docs/credentials.md
davidkretch: Added with links.

adambanker (approved):

Issue #385, opened 2021-02-16
AWS Personalize endpoints


buliin:
When I try to run following code:

```
svc <- personalize(
  config = list(
    credentials = list(
      creds = list(
        access_key_id = "XXX",
        secret_access_key = "XXX"
      )
    ),
    region = "eu-central-1"
  )
)
svcRt <- personalizeruntime(
  config = list(
    credentials = list(
      creds = list(
        access_key_id = "XXX",
        secret_access_key = "XXX"
      )
    ),
    region = "eu-central-1"
  )
)

svcRt$get_recommendations(campaignArn = svc$list_campaigns()$campaigns[[2]]$campaignArn,
                        userId = "999")
```

The campaignArn is taken from list_campaigns function and it returns correct value.
The get_recommendations function ends with the following error:

```
Error in curl::curl_fetch_memory(url, handle = handle) : 
  Could not resolve host: personalizeruntime.eu-central-1.amazonaws.com
```

On the AWS documentation site: https://docs.aws.amazon.com/general/latest/gr/personalize.html
the endpoint for Personalize Runtime is:
personalize-runtime.eu-central-1.amazonaws.com
not the one from error:
personalizeruntime.eu-central-1.amazonaws.com which comes from this function:
https://github.com/paws-r/paws/blob/main/cran/paws.machine.learning/R/personalizeruntime_service.R

Did AWS change the endpoints? Could you fix it in the function?

EDIT: similar situation for personalize-events endpoints

adambanker:
Thank you for submitting this issue. It appears that Paws is currently using the wrong endpoint prefix for the Personalize Runtime service. We'll work on a fix for that.

In the meantime, you can set a custom endpoint when you declare your service object
```r
svcRt <- personalizeruntime(
  config = list(
    credentials = list(
      creds = list(
        access_key_id = "XXX",
        secret_access_key = "XXX"
      )
    ),
    region = "eu-central-1",
    endpoint = "personalize-runtime.eu-central-1.amazonaws.com"
  )
)
```
with the key line being the endpoint at the bottom. This should alleviate your issue until we submit a proper fix to Paws.

davidkretch:
Hello again. Sorry for the delay. A fix for this is now on CRAN. Please let us know if you have any more issues. Thank you!

Issue #62, opened 2018-12-21
Service endpoints not found


davidkretch:
The following packages receive default endpoints that do not exist.

~~paws.chime~~
paws.devicefarm
paws.globalaccelerator
paws.iot1clickdevicesservice
paws.kinesisanalyticsv2
paws.licensemanager
paws.mediastoredata
paws.migrationhub
paws.pinpointemail

davidkretch:
The chime endpoint is `service.chime.amazonaws.com` which appears in the file `endpoints.json` in the Go SDK but I'm not sure how it is derived in the JavaScript SDK.

davidkretch:
We now have a way to manually update endpoints when the one supplied by the service's json file isn't right. I still don't know what the right endpoints for the above services are though.

davidkretch:
I believe this has been fixed.

PR #386, opened 2021-02-16
Uses the provided endpoint prefix rather than the service name

adambanker:
Previously we assumed that the endpoint prefix was the same as the service name. This is usually the case, but this update reads in the endpoint prefix from the api metadata. Addresses #385 and #62


davidkretch (approved):
Looks good to me.

PR #387, opened 2021-02-18
Fix integration tests

davidkretch:
* Fix the integration test make recipe

    Make the integration test recipe run tests on each of the CRAN category packages in the `cran` folder, rather than the legacy paws package in the `paws` folder. The tests in the legacy paws package no longer work because we no longer run the Roxygen2/document process on it.

* Fix duplicated lines in integration test output

    Delete the call to `context("svc")` in generated integration test files. `testthat` no longer needs `context()` and now it only results in a duplicated line in the output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,11 @@ unit: test-common test-codegen
 
 integration:
 	@echo "run integration tests"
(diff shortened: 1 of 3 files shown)


adambanker (approved):
Looks good!

PR #391, opened 2021-02-27
Removes deprecated context call from paws.common and make.paws tests

adambanker:
* testthat suggests that you should no longer use the `context` function because it will automatically use the file name to get the context
* also renames `test-cache.R` in make.paws to `test_cache.R` to follow the underscore convention we have been using


davidkretch (approved):
Looks good to me!

PR #394, opened 2021-03-06
Adds functionality to read in ini files

adambanker:


paws.common/R/iniutil.R
```diff
@@ -0,0 +1,33 @@
...
+# Read in values from an ini file
+read_ini <- function(file_name) {
+  if (!file.exists(file_name)) {
+    stop(sprintf("Unable to find file: %s", file_name))
+  }
+  content <- scan(file_name, what = "", sep = "\n", quiet = T)
+  profiles <- list()
+  current_profile <- ""
+  for (i in 1:length(content)) {
+    if (grepl("\\[.*\\]", content[i])) {
```
davidkretch, line 24: I would make this `"^\\[.*\\]"` to make sure it only captures lines beginning with `[`.

paws.common/R/iniutil.R
```diff
@@ -0,0 +1,33 @@
+# Get the profile name from an ini file
+extract_ini_profile <- function(item) {
+  profile <- gsub("\\[|\\]", "", item)
```
davidkretch, line 3: I would make this `"^\\[.*\\]"` to make sure it only captures lines beginning with `[`.

paws.common/R/iniutil.R
```diff
@@ -0,0 +1,33 @@
...
+  return(profile)
+}
+
+# Get a parameter and its value
+extract_ini_parameter <- function(item) {
+    item_split <- strsplit(item, "=")[[1]]
```
davidkretch, line 9: I think this should split at the first `=`, since this might split into more parts if someone has a `=` in the value.

davidkretch:
We should probably also make keys case insensitive, and ignore lines starting with `;`, going by https://en.wikipedia.org/wiki/INI_file.

adambanker:
> We should probably also make keys case insensitive, and ignore lines starting with `;` or `#`, going by https://en.wikipedia.org/wiki/INI_file.

Can you elaborate on what you were thinking for making it case insensitive? This just returns a list so matching keys would be handled elsewhere

davidkretch:
That's a good point. I was thinking of storing all names as lower case, but that is not actually necessary.

davidkretch:
Looks good to me.

Issue #404, opened 2021-04-27
minio as an s3 alternative not working for most operations


dpastoor:
use of minio https://docs.min.io/docs/minio-quickstart-guide.html for testing or local dev is great to emulate s3, but listing/putting of objects doesn't seem to work

```
library(paws)
svc <- s3(config = list(
  credentials = list(
    creds = list(
      access_key_id = "minioadmin",
      secret_access_key = "minioadmin"
    )
  ),
  endpoint = "http://localhost:9000"
))

svc$list_buckets() # works!

## everything below don't work
#bkt <- svc$create_bucket(Bucket =  "test")

svc$list_objects_v2("test")

svc$put_object(
  Body = "example.R",
  Key = "example.R",
  Bucket = "test"
)
```

all will fail with a consistent: 

```
Error in curl::curl_fetch_memory(url, handle = handle) : 
  Could not resolve host: test.localhost
```

It looks like its prepending the bucketname rather than using slash routing?

Any ideas of how to tweak configuration? or generally where to look. I could likely try to help with a PR if pointed where to look a bit (i'm guessing somewhere in the paws.storage) and definitely testing.

davidkretch:
Huh, sorry about that. It looks like it comes from [a custom S3 service handler for prepending the bucket name to the host name](https://github.com/paws-r/paws/blob/df654f9378d4fe0848edb25254b067c813376e9a/paws.common/R/custom_s3.R#L48). I will take a look this weekend. Sorry for the delay.

For background: Each service uses some set of handlers for building and sending API calls. All of those handlers are in the `paws.common` package. S3 uses the so-called [REST-XML handlers](https://github.com/paws-r/paws/blob/main/paws.common/R/handlers_restxml.R), which is shared with some other APIs, but S3 also has its own custom handlers which do things for S3 in particular, including this one.

If you do put together a PR I'd definitely be grateful, but will try to fix soon in any case.

davidkretch:
Sorry for the delay. I've just submitted a fix. It'll probably be a few days before it gets to CRAN.

One unfortunate side effect: you'll have to change the call to `s3(...)` to include a new argument, `s3_force_path_style`:

```r
library(paws)
svc <- s3(config = list(
  credentials = list(
    creds = list(
      access_key_id = "minioadmin",
      secret_access_key = "minioadmin"
    )
  ),
  endpoint = "http://localhost:9000",
  s3_force_path_style = TRUE
))
```

[...]

PR #407, opened 2021-05-01
Use path style S3 URLs when using custom endpoints or the s3_force_path_style option

davidkretch:
* Use path style URLs for S3 when using custom endpoints, (e.g. localhost:9000/mybucket). This is useful when using software like MinIO. When not using custom endpoints, Paws will in general use "virtual hosted" URLs (e.g. mybucket.s3.amazonaws.com) except in special cases.

* Use path style URLs for S3 when using the new `s3_force_path_style` option when calling `paws::s3()`, e.g.:
    ```r
    s3 <- paws::s3(config = list(
        s3_force_path_style = TRUE
    ))
    ```

Addresses #404.

### DIFF
--- a/paws.common/R/custom_s3.R
+++ b/paws.common/R/custom_s3.R
@@ -65,8 +65,10 @@ update_endpoint_for_s3_config <- function(request) {
 
   if (request$operation$name %in% c("GetBucketLocation")) return(request)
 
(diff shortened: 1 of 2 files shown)


adambanker (approved):
Looks good!

Issue #401, opened 2021-04-03
Functionality to copy s3 file directly to local disk without table intermediate


LBragg:
Hi. Thanks for the great package. It's very comprehensive. 

I am looking for a function equivalent to boto3's download file (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.download_file) , which can also be done via the AWS cli using 'cp' -- copying a file from an s3 bucket to a local file location.

I may have missed something in the docs, any help would be great. Thanks!

adambanker:
Thanks for submitting this issue! Unfortunately, we do not currently have a function like that. The only way we currently support downloads is by retrieving the content in binary format and then writing that to a file as shown here: https://github.com/paws-r/paws/blob/main/examples/s3.R#L66 .

I agree that it would be nice to create a function that supports this use case, so we will put this on our to-do list.

adambanker:
I added the functionality in #408 but it is not available to use yet - but should be soon. That pull request adds the custom `download_file` function to our sdk generator package - [make.paws](https://github.com/paws-r/paws/tree/main/make.paws). The next time we regenerate the paws packages, that function will be added to [paws.storage](https://github.com/paws-r/paws/tree/main/cran/paws.storage), and therefore available to easily install.

If you would like to use the functionality before then, you can pull the latest version of `main`, and regenerate the packages using the makefile - `make install`. However, we were planning on regenerating the packages this week anyway, so it should be available soon without having to do that.

hadley:
#408 doesn't really solve the problem since the data has to travel through R — an ideal solution would be for libcurl (via curl via httr) to save the data directly to disk. (I have a dim recommendation that this is currently hard/impossible to do correctly in httr, but I'm happy to help resolve that.)

adambanker:
That's a good point. We haven't yet looked into a solution that would bypass R, but that does sound better.

I don't believe we have added an argument to a function that is not an api argument. The closest to that is the wrapper function created in #408. As far as other examples of customizations, there are examples of [custom handlers](https://github.com/paws-r/paws/blob/main/paws.common/R/custom_s3.R) here.

hadley:
@DyfanJones I've already forgotten how httr works, but the chances are it'll be much easier with [httr2](https://github.com/r-lib/httr2)

DyfanJones:
> @DyfanJones I've already forgotten how httr works, but the chances are it'll be much easier with [httr2](https://github.com/r-lib/httr2)

[...]

PR #408, opened 2021-05-01
Creates a custom function for downloading files

adambanker:
This is a wrapper function for `s3$get_object` with an additional argument for a filename. This function automatically saves the downloaded object to that location. This is based off of the [boto implementation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.download_file)

Addresses #401


davidkretch:
Sweet. Looks good to me.

Issue #406, opened 2021-04-29
Timeout and retry missing


aleaficionado:
API calls to AWS still timeout after 10 seconds even though the default timeout was changed to null in 0.3.8. I couldn't find a way to change the timeout.

Additionally, it seems that the retry functionality isn't implemented even though the code uses the HTTR lib which provides it. 

I've attached a patch with the changes to paws.common that "fix" the above two issues via the config parameter (added timeout) and implemented in the required functions, but just to point out I have no idea how PAWS or even R are supposed to work.

[network params.txt](https://github.com/paws-r/paws/files/6399602/network.params.txt) This is a diff but gh won't let me upload it as such.

Thanks,
Adi

davidkretch:
Thank you!

Sorry about the issues. This is useful, and in fact we can also make these configurable, like `paws::svc(config = list(max_retries = 5))`.

The issue with timeouts is strange since as you stated the default was updated in the last release and should be no timeout. Would you be willing to try installing the latest paws.common from CRAN with `install.packages("paws.common")`?

In general paws.common gets installed automatically but if you have an older version installed it won't update automatically even if you update the paws package. I think what we'll do is on the next release we'll mandate installing >=0.3.8.

I think we'll want to keep the default as no timeout because some operations download lots of data or otherwise take a long time, which is what prompted the change in 0.3.8.

aleaficionado:
Hi,

First of al great package. I don't think I had an older version of paws.common as I have never used R before submitting this issue, but I will check once I open the work laptop and confirm (we also had this for a couple of devs). To check for timeouts you can try issuing a dummy request to http://example.com:81/ and check the exact timeout you get though you might need to add a dummy service or the like.
My feeling is that some library will set a default timeout of 10 seconds so unless you specify it, it will default to that.

The timeout should not relate to the length of the operation - it's the time between your request being sent and the acknowledgment from the server (and the other way around). So if your timeout is 1 second and you are calling an API that takes 1 minute to complete, the operation will be successful (baring network issues) as long as the server acknowledges your request within one second. You should be able to verify this by setting a timeout of 1 second and downloading or uploading a large file.

Generally you would want a defined timeout and retry as some HTTP requests do get lost (especially with corporate proxies) and you would hang the client with an infinite timeout. E.g. boto3 (or botocore) from AWS uses a timeout of 60 seconds and three retry attempts.

Cheers,
A

davidkretch:
[...]

PR #409, opened 2021-05-02
Support `timeout` and `max_attempts` HTTP request options

davidkretch:
Add support and default values for the following config options:
  - `timeout`: How long to wait in seconds for an acknowledgement of an HTTP request before failing. Default = 60.
  - `max_attempts`: How many times in total to attempt the HTTP request if the request(s) fail. Default = 3.

The `max_attempts` option is named after the [AWS shared configuration item](https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-max_attempts.html).

These are user configurable by adding arguments to the service call, e.g. 

```r
svc(config = list(timeout = 10, max_attempts = 1))
```

Addresses #406. Based on code contributed by @aleaficionado in issue #406.


paws.common/R/net.R
```diff
@@ -90,18 +92,20 @@ issue <- function(http_request) {
   url <- build_url(http_request$url)
   headers <- unlist(http_request$header)
   body <- http_request$body
-  timeout <- if (!is.null(http_request$timeout)) httr::timeout(http_request$timeout)
+  timeout <- if (!is.null(http_request$timeout)) httr::config(connecttimeout = http_request$timeout)
+  times <- if (!is.null(http_request$max_attempts)) http_request$max_attempts else 1
```
adambanker, line 96: This inline if-else statement isn't idiomatic. Maybe we should instead use:
`times <- ifelse(!is.null(http_request$max_attempts), http_request$max_attempts, 1)`

adambanker (commented):
I just had one minor comment. Overall this looks good!

paws.common/R/net.R
adambanker: We also default it to 1 anyway

davidkretch: Addressed in a slightly different way. Let me know what you think. I kept the null guards since we can't guarantee the function will get non-nulls.

adambanker (approved):
Looks good!

Issue #426, opened 2021-06-02
EventBridge functions are not working


Devids:
It looks like eventbridge functions are not working. Every EventBridge function I tried returns the following:
`Error: InvalidSignatureException (HTTP 400). Credential should be scoped to correct : 'events'.`

Simple example for reproduction:
```
eventbridge <- paws::eventbridge()
eventbridge$list_event_buses()
```

I think events should be used as `service_name` instead of eventbridge in `eventbridge_service.R`:

```
.eventbridge$metadata <- list(
   service_name = "eventbridge",
   endpoints...
```

adambanker:
Thanks for submitting this issue! We updated the endpoint earlier this year to use `events` instead of `eventbridge`, but based on what you saw, it looks like we may need to change it somewhere else. I don't think we want to change the `service_name` field, though, since this is for the eventbridge service. I will look more into the issue.

davidkretch:
I'm uploading an updated version of Paws to CRAN where this is working. It should be available this week. Sorry for the delay!

davidkretch:
The fix is now available in the version of paws on CRAN (0.1.12). Sorry again for the delay.

PR #430, opened 2021-06-13
Update CRAN packages for new CRAN rules; fix EventBridge signing name

davidkretch:
* Use the `endpointPrefix` as the signing name when there is no other signing name. This addresses #426.
    + I regenerated Paws and ran all integration tests. EventBridge now works.
* Add a return value section to the documentation explaining what is returned by the service functions (e.g.  `s3()`), as required by CRAN.
* Add quotes around "Amazon Web Services" in all DESCRIPTION files, as required by CRAN.

### DIFF
--- a/make.paws/R/service.R
+++ b/make.paws/R/service.R
@@ -19,6 +19,8 @@ service_file_template <- template(
   #'
   #' ${operations}
   #'
+  #' ${return}
+  #'
   #' @rdname ${service}
   #' @export
   ${service} <- function(config = list()) {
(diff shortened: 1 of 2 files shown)


make.paws/R/service.R
```diff
@@ -199,10 +213,12 @@ protocol_package <- function(api) {
...
 }
 
-# Returns the signing name for an API, either the signing name specified in the
-# API definition or the signing name assigned by `client_config`.
+# Returns the signing name for an API as specified in the API definition. If
+# none exists, return NULL; in this case, the signing name will be inferred at
+# run time.
 signing_name <- function(api) {
   name <- api$metadata$signingName
+  if (is.null(name)) name <- api$metadata$endpointPrefix
```
adambanker, line 221: Oh cool. That's a nice and simple fix

adambanker (approved):
Looks good!

PR #458, opened 2021-09-10
Update custom function for downloading files from s3

DyfanJones:
This method by-passes R when collecting data from s3. Instead it calls `httr:write_disk` from `paws.common:::issue` and relies on `libcurl` to do the heavy lifting through [`curl::curl_fetch_disk`](https://github.com/jeroen/curl/blob/cd292b87269f1a1c960bc849ab9ff43c9ee50e5b/R/fetch.R#L72-L79).

Addresses #401


philiporlando:
Hi, thank you for this great update! I’m trying to confirm my understanding of how the `s3$download_file()` function behaves in terms of memory management in the newer version.

I’m using `s3$download_file()` within an R [{targets}](https://github.com/ropensci-books/targets/) pipeline with large-scale branching targets. I’ve noticed that my system's memory consumption rises significantly during the download process, eventually leading to memory limits being hit and the system crashing. My understanding is that, with this merged PR, `s3$download_file()` should not accumulate memory since files are now written directly to disk instead of being held in memory first. I'm trying to pinpoint where the memory accumulation is happening.

Can you confirm if there are any known memory issues in the current implementation, or if there might still be room for memory to accumulate under certain circumstances? I am troubleshooting my pipeline and would appreciate any insights on whether explicit `gc()` calls might help, or if there’s something I’m missing in how the function behaves.

Thanks in advance!

PR #460, opened 2021-09-29
Update s3.R

alexhallam:
This should be obvious from the context, but I am always forgetting how to read a csv from s3 using paws. Maybe someone else out there feels the same way.


davidkretch:
Hey, sorry for the delay!  Looks good to me.  I added some steps to upload a CSV, and also added examples of reading in the files without writing to disk as suggested by some people in one of the issues.

PR #468, opened 2021-10-29
Obtain credentials via SSO login

### LINKED ISSUE #457: SSO support?
Do y'all have plans to support aws sinle sign on? (e.g. https://docs.aws.amazon.com/singlesignon/latest/userguide/what-is.html)

volkmuth:
Here's an attempt to resolve #457.  This supports getting credentials after a separate sso login (e.g. via `aws sso login --profile my-sso-profile`) and seems to .

If this looks roughly right but you want some refinements let me know and I can try cleaning it up.  For example, I added an `access_token` field to `Creds` which doesn't seem very elegant, though it works.  If the code isn't acceptable please consider it a complicated way of seconding the request in #457 :).


jakegalson:
Thanks for having a go at this - I am interested in this functionality too. Did just run into the following error when trying to run this PR though:

Error in set_config(svc, config) : could not find function "set_config"

volkmuth:
Jake, I tried testing the branch in a container and it worked for me.  Down
below are the steps I followed.  There is a script which runs in the
container and uses sso to access a list of buckets in s3.  Please let me
know if this works for you.  If it doesn't could you please provide more
info?  Or, if you are seeing the problem with a service other than s3
please let me know.  I've tried with ec2, s3 and rds but no others.  If the
test using docker works for you as it does for me, maybe there is something
going wrong in your install in whatever environment you're working in?

Thanks,
Wayne

Dockerfile:
----
from rocker/verse:4.1.1

copy paws /paws

WORKDIR /paws

RUN make deps

RUN make install

WORKDIR /
----
Get copy of repo:

git clone ***@***.***:volkmuth/paws.git
cd paws
git checkout issue457_sso_support
cd ..
build  -t issue457_sso_support:latest .

Create script for testing:
cat > test.R <<__EOT__
#!/usr/bin/env Rscript

library(paws)
Sys.setenv(AWS_PROFILE='my-sso-profile')
foo <- s3()
foo$list_buckets()
__EOT__

Login with sso.  Then run the test in the docker container with:
docker run --rm  -v $HOME/.aws:/root/.aws -v `pwd`:/paws
issue457_sso_support /paws/test.R

jakegalson:
Thanks for this - I will have a go and let you know if it works.

jakegalson:
(I was using ***@***.***”), which may have been the problem)

jakegalson:
It still doesn't work for me using the container:

> foo$list_buckets()
Error in if (grepl("(^;)|(^#)", content[i])) { : 
  argument is of length zero

Presuming it must be something to do with the format of my aws config file

jakegalson:
A big +1 from me. All working nicely now.

davidkretch:
Hey, thank you for all this! I'll try it out this weekend.

volkmuth:
@davidkretch I updated the pull request to fix a couple of issues.  I ran the unit-tests workflow on my fork and it fails.  It looks like the failure occurs because paws.security.identity is installed from cran rather than from the local, updated code.

I notice that the unit-tests workflow failed for the 0.1.12 release https://github.com/paws-r/paws/actions/runs/1155680322.

Is the workflow expected to pass?

davidkretch:
Yeah, it should pass at least locally. I'm not sure why it failed for that version; could have been code coverage even. I wouldn't worry about it for now. I'll take a look tomorrow.

davidkretch:
Oh, I see what you mean now. This looks good to me. Thank you!

Issue #482, opened 2022-01-31
linux os assumption for sso_credential_process


yonicd:
https://github.com/paws-r/paws/blob/9bf12c3ce55ac10b9f88f428d1b1e41b751b001a/paws.common/R/credential_providers.R#L174

This assumes linux style path for looking up the sso config file, which creates an error in winos.

```r
sso_credential_process(sso_start_url, sso_account_id, sso_region, sso_role_name)
Error: lexical error: invalid char in json text.
                                       ~/.aws/sso/cache/ef83fd07ea6f19
                     (right here) ------^
```

this works 

```r
root <- ifelse(Sys.info()[[1]]=='Windows',Sys.getenv('HOMEPATH'),'~')
sso_cache <- file.path(root, ".aws", "sso", "cache", json_file)
```

dpastoor:
prob could just simplify to calling `normalizePath()` which then would expand out the `~`

if you go down the "custom location based on env", would prob want to use the same env that the aws sdk is going to look which is $HOME or $USERPROFILE - https://docs.aws.amazon.com/sdkref/latest/guide/file-location.html preferentially

yonicd:
normalize path doesnt work always for winos

```r
normalizePath('~')
[1] "C:\\Users\\jonathan.sidi\\OneDrive - Sage Therapeutics\\Documents"

Sys.getenv('HOMEPATH')
[1] "\\Users\\jonathan.sidi"
```

dpastoor:
hmm - interesting - the good ol R funkiness of user home dir default oh windows - does $HOME give back the same as USERPROFILE or the normalizedPath?

dpastoor:
ah I knew I remember this coming up elsewhere - https://github.com/r-lib/gert/issues/35

yonicd:
`HOME` gives back `normalizePath('~')` in winos

yonicd:
I got to HOMEPATH from this [thread](https://superuser.com/a/168719)

davidkretch:
Hello, sorry for the delay.  The fix you submitted is now on CRAN.  You can get it by downloading the latest `paws.common`.  Thank you for the PR!!

PR #483, opened 2022-01-31
fix sso credential process in winos

yonicd:
see issue https://github.com/paws-r/paws/issues/482


davidkretch:
Looks good to me.

PR #494, opened 2022-03-18
fix params without values

tomlue:
Some AWS endpoints take parameters without values (e.g `paws::s3()$get_bucket_website`)

This PR fixes the below error and other endpoints with valueless parameters
```R
s3$get_bucket_website("some bucket")
Error: InvalidArgument (HTTP 400). The website parameter must not have a value
```


davidkretch:
Looks good to me.

davidkretch:
Thank you!!

davidkretch:
This is on its way to CRAN as paws.common 0.3.17.  Thank you for the fix!

Issue #501, opened 2022-06-28
Issue: upload_part_copy return list with empty elements


DyfanJones:
It looks like `upload_part_copy` is returning back empty elements. 

```r
size = 1e6
df = data.frame(
  var1 = sample(letters, size, TRUE),
  var2 = 1:size
)
write.csv(df, "temp.csv")

s3_client = paws.storage::s3()

bucket = "my-bucket"
src_key = "file.csv"

s3_client$put_object(
  Bucket = bucket,
  Key = src_key,
  Body = "temp.csv"
)

file_size = s3_client$head_object(
  Bucket = bucket,
  Key = src_key
)$ContentLength

dest_key = "new_file.csv"

src = sprintf("/%s/%s", bucket, src_key)

upload_id = s3_client$create_multipart_upload(
  Bucket = bucket, Key = dest_key
)$UploadId

start = seq(0, file_size-1, 10 * (2 ** 20))
end = c(start[-1]-1, file_size-1)
chunks = sprintf("bytes=%s-%s", start, end)

resp = lapply(seq_along(chunks), function(i) {
  s3_client$upload_part_copy(
    CopySource = src,
    Bucket = bucket,
    Key = dest_key,
    UploadId = upload_id,
    PartNumber = i,
    CopySourceRange =  chunks[i]
  )
})

resp
# [[1]]
# [[1]]$CopySourceVersionId
# character(0)
# 
# [[1]]$CopyPartResult
# [[1]]$CopyPartResult$ETag
# character(0)
# 
# [[1]]$CopyPartResult$LastModified
# POSIXct of length 0
# 
# 
# [[1]]$ServerSideEncryption
# character(0)
# 
# [[1]]$SSECustomerAlgorithm
# character(0)
# 
# [[1]]$SSECustomerKeyMD5
# character(0)
# 
# [[1]]$SSEKMSKeyId
# character(0)
# 
# [[1]]$BucketKeyEnabled
# logical(0)
# 
# [[1]]$RequestCharged
# character(0)
# 
# 
# [[2]]
# [[2]]$CopySourceVersionId
# character(0)
# 
# [[2]]$CopyPartResult
# [[2]]$CopyPartResult$ETag
# character(0)
# 
# [[2]]$CopyPartResult$LastModified
# POSIXct of length 0
# 
# 
# [[2]]$ServerSideEncryption
# character(0)
# 
# [[2]]$SSECustomerAlgorithm
# character(0)
# 
[...]

PR #503, opened 2022-06-29
Bug: unmarshal xml objects when elements are at root

DyfanJones:
Helps to ensure unmarshal xml http data response when elements are at root

Addresses speed issue from ticket: https://github.com/paws-r/paws/issues/501


davidkretch:
This is going to take a little longer answer about the build process and I'll put together a document this week. This is a little tricky in that files in `cran`, e.g. `cran/paws.storage/R/s3_operations.R` are all generated automatically. The way you build them is by running `make build` from the command line which runs the makefile build recipe. 

I think we have some other ways of handling special situations like this:
* https://github.com/paws-r/paws/blob/main/paws.common/R/custom_s3.R, for custom handlers that get run on only S3 requests but are not themselves user accessible functions <-- this is where I think it may go...
* https://github.com/paws-r/paws/blob/main/make.paws/R/custom/s3.R, for custom user accessible functions. 

This is a weird one. S3 was probably one of their first APIs and is probably why you see weird stuff like this.

DyfanJones:
I think I have thought of a way to allow the operation name to be modify without requiring a seperate custom function. If we override the operation_name within the function: `make_operation` https://github.com/paws-r/paws/blob/84e5fc9d8dc05ae40c63f3935bee82256e414a69/make.paws/R/operations.R#L48-L62

To do this we can have a yaml file with a list of operations we need to override for example:

```
---
- name: UploadPartCopy
  override: CopyPart
```

Then we can have a simple function that checks the yaml file and replaces them accordingly:

```r
operation_name <- function(operation_name){
  path <- system_file("extdata/operation_name_override.yml", package = methods::getPackageName())
  out <- yaml::read_yaml(path)
  found <- vapply(out, function(x) x$name == operation_name, FUN.VALUE = logical(1))
  override <- NULL
  if(any(found))
    override <- out[[which(found)]]$override
  operation_name <- if(!is.null(override)) override else operation_name
  return(quoted(operation_name))
}
```

This can be fed into `make_operation`:

```r
# Make an operation, both function and documentation.
make_operation <- function(operation, api) {
  render(
    operation_template,
    docs = make_docs(operation, api),
    service = package_name(api),
    operation_name = operation_name(operation$name),
    function_name = get_operation_name(operation),
    params = get_operation_params(get_operation_input_shape(operation, api)),
    operation_input = get_operation_input(operation, api),
    operation_output = get_operation_output(operation),
    http_method = quoted(operation$http$method),
    http_path = quoted(operation$http$requestUri)
  )
}
```

This way the operation method is kept up to date with any changes in future builds.

DyfanJones:
Ah sorry the `make build` has modified so many files :S

DyfanJones:
it looks like alot of the url links are breaking:
```
#' regions and endpoints, see [Regions and
#' Endpoints](https://docs.aws.amazon.com/general/latest/gr/rande.html#cloudsearch_region).
```

```
#' regions and endpoints, see <a
#' href="https://docs.aws.amazon.com/general/latest/gr/rande.html#cloudsearch_region"
#' target="_blank">Regions and Endpoints</a>.
```

not sure why this is the case

davidkretch:
This looks good. I would make the following changes:
1. Separate out the code re-generation into a separate PR to be easier to review. 
2. Add tests if possible.
3. Break lines at 80 characters, the current limit in the code base. Unless you want to change the limit, but as a code readability aid I think it should be between 80-100.

DyfanJones:
@davidkretch, thanks for all your help with this PR. It has really helped me get up to speed with the maintance of Paws. 

Made amendments to the PR following your request.

> This looks good. I would make the following changes:
> 
>     1. Separate out the code re-generation into a separate PR to be easier to review.
> 
>     2. Add tests if possible.
> 
>     3. Break lines at 80 characters, the current limit in the code base. Unless you want to change the limit, but as a code readability aid I think it should be between 80-100. Unfortunately we don't have any linter or auto-formatter for something like this.

When you have time do you mind reviewing the PR.

Many thanks for all your help :)

davidkretch:
Looks good! Thanks!

davidkretch:
Feel free to merge in.

Issue #510, opened 2022-07-18
s3$put_bucket_lifecycle_configuration is not accepting empty prefix


y-fischer:
I really like the package and use it for my daily work!

But I realized, that it is not possible to put a lifecycle on a bucket with an empty prefix like:
`s3$put_bucket_lifecycle_configuration(Bucket = "testbucket",
                                      LifecycleConfiguration = list(Rules = list(list(ID = "five-days",
                                                                                      Filter = list(Prefix = ""),
                                                                                      Expiration = list(Days = 5),
                                                                                      Status = "Enabled"))))`
Error message:
`Fehler: MalformedXML (HTTP 400). The XML you provided was not well-formed or did not validate against our published schema`

Amazon states (https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html):
> You can specify an empty filter, in which case the rule applies to all objects in the bucket.

How can I put a lifecycle configuration that is affecting all objects in the bucket?
Thanks for any help or improvement!

DyfanJones:
Hi @y-fischer  thanks for raising the issue, it looks like a possibly xml build issue. I will have a look to see if i can get it in the next release 😄

y-fischer:
That would be great! Thanks!

DyfanJones:
This has now been fixed in PR #511. We are getting ready for the next cran release, but in the meantime you can get the dev version through:

```r
remotes::install_github(repo = "paws-r/paws/paws.common/")
```

DyfanJones:
paws.common 0.4.0 is now on the cran and should resolve this issue :)

I will close this ticket for now. If this issue hasn't been resolved for you please feel free to re-open.

y-fischer:
Everything works as expected. Thanks a lot!

PR #511, opened 2022-07-22
allow xml_build to accept empty characters

DyfanJones:
This PR address #510 and allows empty character strings i.e. `""` to be accepted when building the xml to send to AWS.


DyfanJones:
🤔  not sure why devtools is not avialable:

`Error: Package required but not available: ‘devtools’`

Will try the unit tests again on monday. Side note it worked all ok on local :D

davidkretch:
Looks good to me. Sorry for the delay. I re-ran the CI jobs.

DyfanJones:
From looking at the raw logs it looks like  textshaping is the issue:

```
2022-07-27T10:07:09.0158735Z -------------------------- [ERROR MESSAGE] ---------------------------
2022-07-27T10:07:09.0161796Z <stdin>:1:10: fatal error: hb-ft.h: No such file or directory
2022-07-27T10:07:09.0162207Z compilation terminated.
2022-07-27T10:07:09.0162815Z --------------------------------------------------------------------
2022-07-27T10:07:09.0169503Z ERROR: configuration failed for package ‘textshaping’
```

I think we could possible fix this by adding some more binaries suggested in this issue: https://github.com/r-lib/textshaping/issues/21

PR #516, opened 2022-08-01
set perl parameter to TRUE

DyfanJones:
This is a minor change so that the encoders can take advantage of the up coming improvements to `grep`, `sub`, `gsub` and `strsplit`

> The performance of grep, sub, gsub and strsplit has been improved, particularly with perl = TRUE and fixed = TRUE. Use of useBytes = TRUE for performance reasons should no longer be needed and is discouraged. It may lead to incorrect results. 
https://cran.r-project.org/doc/manuals/r-devel/NEWS.html


DyfanJones:
Improved `paws_url_decoder` method. This helps when we scale up and decode large strings, for example:

``` r
# decode encoded url strings
paws_url_decoder <- function(URL) {
  vapply(URL, function(url){
    # split string into separate characters
    chars <- strsplit(url, "", perl = TRUE)[[1]]
    
    # locate % position
    found <- grep("%", chars, fixed = TRUE)
    
    if (length(found)) {
      # get position of encoded part
      encoded_pt = sort.int(c(found + 1, found + 2))
      encoded_char = paste0(chars[encoded_pt], collapse = "")
      
      # get raw vector of encoded parts (character form)
      # for example: "%20" -> "20"
      encoded_char_len = nchar(encoded_char)
      encoded <- substring(
        encoded_char,
        seq(1, encoded_char_len, 2),
        seq(2, encoded_char_len, 2)
      )
      
      # update character % position
      found <- grep("%", chars[-encoded_pt], fixed = TRUE)
      
      # convert split url to raw
      char_raw <- charToRaw(paste(chars[-encoded_pt], collapse=""))
      
      # replace character % with decoded parts
      char_raw[found] <- as.raw(as.hexmode(encoded))
      
      return(rawToChar(char_raw))
    }
    return(url)
  }, character(1),  USE.NAMES = FALSE)
}


string <- "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._~`!@#$%^&*()=+[{]}\\|;:'\",<>/? "
pattern <- "[^ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._~-]"

url = paste(sample(strsplit(string, "")[[1]], 1e7, replace = T), collapse = "")
url_encode = paws.common:::paws_url_encoder(url, pattern)
bm = bench::mark(
  old_method = paws.common:::paws_url_decoder(url_encode),
  new_method = paws_url_decoder(url_encode),
  iterations = 20
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
bm |> ggplot2::autoplot()
#> Loading required namespace: tidyr
```



<sup>Created on 2022-08-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This method is more scalable as a `vapply` has been removed. This means `paste0` doesn't need to be called from the `vapply` causing `gc()` not to be called as  much.

DyfanJones:
Feel a little silly now :P just realised we could use curl's `curl_unescape` when decoding url's. As `httr` depends on `curl` we aren't increasing our overall dependencies. Plus there is a bonus of us not needing to maintain a decoder :P.

DyfanJones:
Sadly we can't use `curl_escape` as it is fixed in what it won't escape. In python `urllib.parse.quote` there is a parameter `safe` that allows user to pass extra characters that shouldn't be escaped (https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote). I might ask `curl` to see if this would be an option or if it is our of scope for that package :) in the mean time we can use the `paws_url_encoder`

DyfanJones:
Hmmm, not sure why make.paws is failing. @davidkretch do you have any ideas why this could be. Error below:

```
convert(text) not equal to `expected`.
1/1 mismatches
x[1]: "foo"
y[1]: "[foo](https://httpbin.org/anything#foo?bar=baz)"
```

DyfanJones:
Issue seems fixed for now. Will keep an eye on it just incase.

PR #518, opened 2022-08-09
use cpp for url encoding

DyfanJones:


paws.common/DESCRIPTION
```diff
@@ -23,6 +25,7 @@ Imports:
     jsonlite,
     methods,
     utils,
+    Rcpp,
```
davidkretch, line 28: I think my biggest question is does Rcpp need to be an import, or can it be a build time dependency (whatever that means in R)? I think for Windows users in particular, Rcpp is a lot of extra stuff to download/install.

For example dplyr which definitely uses C++ does not have Rcpp as an import and I would like to figure out how they do it.

DyfanJones: That is a good question. I will have a look.

davidkretch: According to https://cran.r-project.org/web/packages/Rcpp/vignettes/Rcpp-package.pdf, Rcpp is required both in Imports and LinkingTo. 

dplyr doesn't have it because it got rid of Rcpp in version 1.0.0 and replaced it with `vctrs` which uses C and I guess R standard library instead of C++ and Rcpp.

I think we'll need to figure out the impact on build time and on Windows users though.

DyfanJones: I think this stackoverflow answers some of our concerns. https://stackoverflow.com/questions/41434895/does-installing-an-r-package-with-rcpp-require-rtools-windows

Once cran has created the binaries windows computers can install the package without the need for Rtools. However if a user tries to install the package from source then Rtools are required.

From my understanding there is alot of packages that utilises rcpp nowadays. There is another c++ interface for R cpp11 which offers similar support to Rcpp. However I believe rtools will be required when installing from source.

paws.common/src/encoding.cpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  //create an output string
+  std::string output = "";
+  //For each character..
+  for(int i=0; i < (signed) url.length(); i++){
+
+    //If it's in the list of reserved ones, just pass it through
+    if (unreserved_chars_map.find(url[i]) != unreserved_chars_map.end()) {
+      output.append(&url[i], 1);
```
davidkretch, line 52: The `&url[i]` is weird because it means address of the i-th element of `url`.

paws.common/src/encoding.cpp
```diff
@@ -0,0 +1,111 @@
+#include <Rcpp.h>
+using namespace Rcpp;
+
+// This encoder has been modified from the excellent urltools R package
+// https://github.com/Ironholds/urltools/blob/master/src/encoding.cpp
+
+std::string to_hex(char x){
+
+  //Holding objects and output
```
davidkretch, line 9: Nit: I feel like there should be a shorter, easier to read way to convert using `sprintf` or something but I'm not able to get it easily so not really important.

paws.common/src/encoding.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  //For each string..
+  for (int i = 0; i < input_size; ++i){
+
+    //Check for user interrupts.
+    if((i % 10000) == 0){
+      Rcpp::checkUserInterrupt();
+    }
+
+    if(urls[i] == NA_STRING){
```
davidkretch, line 88: Nit: I feel like this should be a `continue` without an else to avoid the extra level of indentation.
```
if (urls[i] == NA_STRING) {
  output[i] = NA_STRING;
  continue;
}
<no else>

**Comment by dpastoor** [comment]
hey guys - I know this is closed, but I'd just want to chime in that given how small this implementation is, I'd def suggest going with cpp11 instead of Rcpp - This is listed out in the motivations https://cpp11.r-lib.org/articles/motivations.html#header-only

but simply - it should be lighter, faster, and given its header only it will avoid any chance at problems for linking issues when upgrading versions.

Realistically, this might not have a significant impact, but as you might move other code hot paths into cpp could get out in front of this now.

Just my two cents!

File: paws.common/DESCRIPTION (hunk 1 above)
Comment (dpastoor) [reply paws.common/DESCRIPTION:28]: as an FYI - dplyr _used_ to depend on Rcpp and BH but before 1.0 made a huge push to remove all need for that with their own implementation + stuff out of vctrs package to improve compilation performance and more tightly control the dep tree

**Comment by DyfanJones** [comment]
Hi @dpastoor thanks for letting us know. I am still fairly new in adding cpp into R packages. I have done some initial benchmarks around rcpp and cpp11 performance of the encoder. To do this I have converted the existing implementation https://github.com/DyfanJones/EncodeCpp11 (I am not sure if I have added any overhead to this).

Here is the initial benchmark.

``` r
library(ggplot2)
string <- "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._~`!@#$%^&*()=+[{]}\\|;:'\",<>? "
iter <- 1:7

# Note: python is at a disadvantage here as it needs to convert it's output from python to R.
parse <- reticulate::import("urllib.parse")

bm <- list()
for (i in iter) {
  url <- paste(sample(strsplit(string, "")[[1]], (10 ^ i), replace = T), collapse = "")
  bm[[i]] <- bench::mark(
    url_encode_rcpp = paws.common:::paws_url_encoder(url),
    url_encode_cpp11 = EncodeCpp11::cpp11_url_encoder(url),
    url_encode_curl = curl::curl_escape(url),
    url_encode_python = parse$quote(url),
    iterations = 100
  )
}

result <- do.call(rbind, bm)
result$size <- sort(rep(10 ^ iter, 4))
result$expression <- as.character(result$expression)

res <- tidyr::unnest(result, c(time, gc))
ggplot(res, aes(time, expression)) +
  geom_violin() +
  facet_wrap("size", scales = "free_x") +
  theme(
    # text = element_text(size = 8),
    axis.text.x = element_text(angle = 20, vjust = 0.5, hjust=1)
  )
```



<sup>Created on 2022-09-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I am not seeing any noticeable performance improvements. I think there is a slight performance decrease, however this could be my conversion (from Rcpp to cpp11) and not cpp11 itself. From my understanding of cpp11 has compile time and memory improvements over Rcpp.

I do like to keep paws as light as possible.

Full benchmarks can be found here: https://github.com/DyfanJones/EncodeCpp11/blob/main/inst/
Practise package (EncodeCpp11) for cpp11 encoder: https://github.com/DyfanJones/EncodeCpp11

PR #523, opened 2022-08-17
Remove Unicode line terminators from documentation

davidkretch:
Fix LaTeX errors that appeared in the last paws.compute, caused by non-printable Unicode line terminator U+2028 copied from the AWS documentation, with error message:

! LaTeX Error: Unicode character   (U+2028)
               not set up for use with LaTeX.

See https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/paws.compute-00check.html

Test plan:
1. Regenerate Paws: `make build`.
2. Check for line terminator. `grep $'\u2028' cran` (no results):
```sh
Davids-MacBook-Air:paws davidkretch$ grep -r $'\u2028' cran
Davids-MacBook-Air:paws davidkretch$ 
```


DyfanJones:
This looks all good to me :)

Issue #537, opened 2022-09-19
select_object_content not recognizing InputSerialization parameter


dleopold:
I am trying to use select_object_content to query some data in an s3 bucket, but I am getting this error:

`Error: MissingRequiredParameter (HTTP 400). InputSerialization is required. Please check the service documentation and try again.`

despite the fact that I am providing the InputSerialization. This is the code I am running:

```
svc <- s3()

svc$select_object_content(
  Bucket = "arrow-testing",
  Key = "iris-data/part-0.parquet",
  Expression = "SELECT * from s3object",
  ExpressionType = "SQL",
  InputSerialization = list('Parquet' = list()),
  OutputSerialization = list('CSV' = list( FieldDelimiter = "," ))
)
```
The same query run using the AWS cli on the same machine works as expected:

```
aws s3api select-object-content \
--bucket arrow-testing \
--key 'iris-data/part-0.parquet' \
--expression "SELECT * FROM s3object s where s.\"Species\" = 'setosa'" \
--expression-type 'SQL' \
--input-serialization '{"Parquet": {}}' \
--output-serialization '{"CSV": {"FieldDelimiter": ","}}' "OutputFile"
```
Here is the sesson info:

```
> sessionInfo()
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Pop!_OS 21.04

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8    LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] paws_0.1.12

loaded via a namespace (and not attached):
 [1] arrow_7.0.0         digest_0.6.29       assertthat_0.2.1    paws.common_0.3.15  R6_2.5.1            magrittr_2.0.3     
[...]

PR #540, opened 2022-09-26
Bug empty list

DyfanJones:
Allow empty lists to be parse from service parameters.

This PR address ticket: https://github.com/paws-r/paws/issues/537


davidkretch:
I think the code looks good but I think it could really use a test to demonstrate what case it's addressing.

DyfanJones:
Sounds good to me, i will create the unit test

Issue #441, opened 2021-08-13
Support IMDSv2


davidkretch:
See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

achimgaedke:
DevOps guided by "AWS Foundational Security Best Practices standard" would love to see this:

https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#fsbp-ec2-8

I'm considering working around it with reticulate and boto3.

DyfanJones:
Closing this as paws.common 0.5.2 supports this :)

PR #552, opened 2022-11-06
Support IMDSv2

jornfranke:
Support IDMSv2.

Tested with Sagemaker Notebooks with IMDSv2 enforced.

Configuration:
~~Optionally, set the environment variable "PAWS_EC2_IMDSV2_TOKEN_TIMEOUT_SECONDS" to the number of seconds after which the token should expire. The default (30 seconds) should be sufficient for nearly all use cases~~ no configuration we use the same method as in boto3

Solves: https://github.com/paws-r/paws/issues/441

### DIFF
--- a/paws.common/R/config.R
+++ b/paws.common/R/config.R
@@ -208,8 +208,10 @@ get_instance_metadata <- function(query_path = "") {
       NULL
     }
   )
-  if (!(is.null(metadata_token_response) && metadata_token_response$status_code != 200)) {
-      token=rawToChar(metadata_token_response["body"])
+  if (!(is.null(metadata_token_response)) && metadata_token_response$status_code == 200) {
+      if (length(metadata_token_response["body"])>0) {
+          token=rawToChar(metadata_token_response["body"])
+      } 
   }
   metadata_url <- file.path(
     "http://169.254.169.254/latest/meta-data",
--- a/paws.common/R/config.R
+++ b/paws.common/R/config.R
@@ -193,7 +193,12 @@ get_instance_metadata <- function(query_path = "") {
     "http://169.254.169.254/latest/api/token"
   )
  metadata_token_request <-
-    new_http_request("PUT", metadata_token_url, timeout = 1, header=c("X-aws-ec2-metadata-token-ttl-seconds"= tokentimeout))
+    new_http_request(
+        "PUT", 
+        metadata_token_url, 
+        timeout = 1, 
+        header=c("X-aws-ec2-metadata-token-ttl-seconds"= tokentimeout)
+    )
 
   metadata_token_response <- tryCatch(
     {
@@ -203,16 +208,21 @@ get_instance_metadata <- function(query_path = "") {
       NULL
     }
   )
-  if (!((is.null(metadata_token_response) && metadata_token_response$status_code != 200))) {
+  if (!(is.null(metadata_token_response) && metadata_token_response$status_code != 200)) {
       token=rawToChar(metadata_token_response["body"])
   }
   metadata_url <- file.path(
     "http://169.254.169.254/latest/meta-data",
     query_path
   )
-  if (!(token=="")) {
+  if (token!="") {
   metadata_request <-
-    new_http_request("GET", metadata_url, timeout = 1, header=c("X-aws-ec2-metadata-token"= token))
+    new_http_request(
+        "GET", 
+        metadata_url, 
+        timeout = 1, 
+        header=c("X-aws-ec2-metadata-token"= token)
+    )
   } else { # use IMDSv1 in case IMDSv2 is not available - not recommended - very insecure
   metadata_request <-
     new_http_request("GET", metadata_url, timeout = 1)
--- a/paws.common/R/config.R
+++ b/paws.common/R/config.R
@@ -175,7 +175,8 @@ get_profile_name <- function(profile = "") {
   return(profile)
 }
 
-# Gets the instance metadata by making an http request.
+# Gets the instance metadata by making an http request to an instance metadata services
+# Please see security recommendations by AWS: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
 get_instance_metadata <- function(query_path = "") {
   token_ttl <- '21600' # same approach as in boto3: https://github.com/boto/botocore/blob/master/botocore/utils.py#L376
   # Do not get metadata when the disabled setting is on.
@@ -221,7 +222,7 @@ get_instance_metadata <- function(query_path = "") {
         timeout = 1, 
         header=c("X-aws-ec2-metadata-token"= token)
     )
-  } else { # use IMDSv1 in case IMDSv2 is not available - not recommended - very insecure
+  } else {
   metadata_request <-
     new_http_request("GET", metadata_url, timeout = 1)
   }
--- a/paws.common/R/config.R
+++ b/paws.common/R/config.R
@@ -177,16 +177,12 @@ get_profile_name <- function(profile = "") {
 
 # Gets the instance metadata by making an http request.
 get_instance_metadata <- function(query_path = "") {
-
+  token_ttl <- '21600' # same approach as in boto3: https://github.com/boto/botocore/blob/master/botocore/utils.py#L376
   # Do not get metadata when the disabled setting is on.
   if (trimws(tolower(get_env("AWS_EC2_METADATA_DISABLED"))) %in% c("true", "1")) {
     return(NULL)
   }
   # Get token timeout for IMDSv2 tokens
-  tokentimeout=trimws(tolower(get_env("PAWS_EC2_IMDSV2_TOKEN_TIMEOUT_SECONDS")))
-  if (is.na(as.numeric(tokentimeout))) {
-        tokentimeout="30"
-  }
   token <-  "" # Token to be used in case of more secure IMDSv2 authentication
   #try IMDSv2  (more information): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
    metadata_token_url <-  file.path(
@@ -197,7 +193,7 @@ get_instance_metadata <- function(query_path = "") {
         "PUT", 
         metadata_token_url, 
         timeout = 1, 
-        header=c("X-aws-ec2-metadata-token-ttl-seconds"= tokentimeout)
+        header=c("X-aws-ec2-metadata-token-ttl-seconds"= token_ttl)
     )
 
   metadata_token_response <- tryCatch(
--- a/paws.common/R/config.R
+++ b/paws.common/R/config.R
@@ -182,14 +182,41 @@ get_instance_metadata <- function(query_path = "") {
   if (trimws(tolower(get_env("AWS_EC2_METADATA_DISABLED"))) %in% c("true", "1")) {
     return(NULL)
   }
+  # Get token timeout for IMDSv2 tokens
+  tokentimeout=trimws(tolower(get_env("PAWS_EC2_IMDSV2_TOKEN_TIMEOUT")))
+  if (is.na(as.numeric(tokentimeout))) {
+        tokentimeout="30"
+  }
+  token <-  "" # Token to be used in case of more secure IMDSv2 authentication
+  #try IMDSv2  (more information): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+   metadata_token_url <-  file.path(
+    "http://169.254.169.254/latest/api/token"
+  )
+ metadata_token_request <-
+    new_http_request("PUT", metadata_token_url, timeout = 1, header=c("X-aws-ec2-metadata-token-ttl-seconds"= tokentimeout))
 
+  metadata_token_response <- tryCatch(
+    {
+      issue(metadata_token_request)
+    },
+    error = function(e) {
+      NULL
+    }
+  )
+  if (!((is.null(metadata_token_response) && metadata_token_response$status_code != 200))) {
+      token=rawToChar(metadata_token_response["body"])
+  }
   metadata_url <- file.path(
     "http://169.254.169.254/latest/meta-data",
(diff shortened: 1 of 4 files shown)


DyfanJones:
Hi @jornfranke thanks for the PR, can you create a unit test for this.

paws.common/R/config.R
```diff
@@ -182,14 +182,41 @@ get_instance_metadata <- function(query_path = "") {
...
+    }
+  )
+  if (!((is.null(metadata_token_response) && metadata_token_response$status_code != 200))) {
+      token=rawToChar(metadata_token_response["body"])
+  }
   metadata_url <- file.path(
     "http://169.254.169.254/latest/meta-data",
     query_path
   )
+  if (!(token=="")) {
```
DyfanJones, line 213: Shouldn't we use `token != ""` ?

paws.common/R/config.R
```diff
@@ -182,14 +182,41 @@ get_instance_metadata <- function(query_path = "") {
...
+  if (is.na(as.numeric(tokentimeout))) {
+        tokentimeout="30"
+  }
+  token <-  "" # Token to be used in case of more secure IMDSv2 authentication
+  #try IMDSv2  (more information): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+   metadata_token_url <-  file.path(
+    "http://169.254.169.254/latest/api/token"
+  )
+ metadata_token_request <-
+    new_http_request("PUT", metadata_token_url, timeout = 1, header=c("X-aws-ec2-metadata-token-ttl-seconds"= tokentimeout))
```
DyfanJones, line 196: Doesn't this exceed 80 characters?  We haven't set up an lint just yet, however if you could update this to multiple lines that would be perfect :D something like

```
 metadata_token_request <- new_http_request(
    "PUT", 
    metadata_token_url,
    timeout = 1,
    header=c("X-aws-ec2-metadata-token-ttl-seconds"= tokentimeout)
)
```

paws.common/R/config.R
```diff
@@ -182,14 +182,41 @@ get_instance_metadata <- function(query_path = "") {
...
+  # Get token timeout for IMDSv2 tokens
+  tokentimeout=trimws(tolower(get_env("PAWS_EC2_IMDSV2_TOKEN_TIMEOUT_SECONDS")))
+  if (is.na(as.numeric(tokentimeout))) {
+        tokentimeout="30"
+  }
+  token <-  "" # Token to be used in case of more secure IMDSv2 authentication
+  #try IMDSv2  (more information): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+   metadata_token_url <-  file.path(
+    "http://169.254.169.254/latest/api/token"
+  )
```
DyfanJones, line 194: small thing, do we need file.path for just assigning a character :)

paws.common/R/config.R
```diff
@@ -182,14 +182,41 @@ get_instance_metadata <- function(query_path = "") {
...
 
+  metadata_token_response <- tryCatch(
+    {
+      issue(metadata_token_request)
+    },
+    error = function(e) {
+      NULL
+    }
+  )
+  if (!((is.null(metadata_token_response) && metadata_token_response$status_code != 200))) {
```
DyfanJones, line 206: Another small thing, isn't there 1 too many brackets? wouldn't something like this do?

```
if (!(is.null(metadata_token_response) && metadata_token_response$status_code != 200))
```

paws.common/R/net.R
```diff
@@ -55,7 +55,7 @@ HttpResponse <- struct(
 # @param close Whether to immediately close the connection, or else reuse connections.
 # @param timeout How long to wait for an initial response.
 # @param dest Control where the response body is written
-new_http_request <- function(method, url, body = NULL, close = FALSE, timeout = NULL, dest = NULL) {
+new_http_request <- function(method, url, body = NULL, close = FALSE, timeout = NULL, dest = NULL, header = list()) {
```
DyfanJones, line 58: I believe we are missing the corresponding documentation for header

```
@param header blah blah
```

Do you mind updating the documentation for this? I believe you will need to run roxygen2 to create the doc for it.

DyfanJones:
@davidkretch when you have time do you have anything extra to add to this PR?

jornfranke:
thanks a lot for the detailed and fast review. All valid, will adapt it as soon as possible.
About the unit test: It would be merely an integration test (in the end just connecting to IMDSv2). I have to check if there is a previous test for the IMDSv1 and then copy this to create one for IMDSv2.

paws.common/R/config.R
jornfranke: This was taken over from the previous function - I just kept it the same to avoid as little inconsistencies as possible.

paws.common/R/config.R
jornfranke: yes you are right, it is a mistake

paws.common/R/config.R
jornfranke: will do

paws.common/R/config.R
jornfranke: will adapt it

paws.common/R/net.R
jornfranke: yes of course - this one slipped through

jornfranke:
> Hi @jornfranke thanks for the PR, can you create a unit test for this.

It makes sense, but we do merely integration with the new IMDSv2 API. So we would need to create a mock IMDSv2 service and then write the test cases. I have not found one test for the previous IMDSv1 call (https://github.com/paws-r/paws/blob/main/paws.common/tests/testthat/test_config.R). Can you point me to one?

Otherwise I propose to create a new issue for the integration test of IMDS (for both versions) as this probably is a bigger effort than providing the feature. Nevertheless, maybe you have already integration tests for other HTTP services used in paws? 

For now I did only test manually the integration using AWS Sagemaker Notebooks with IMDSv1 enforced. The library without the change does not work there. Only after adding IMDSv2 supprot it works.

Let me know.

DyfanJones:
> > Hi @jornfranke thanks for the PR, can you create a unit test for this.
> 
> It makes sense, but we do merely integration with the new IMDSv2 API. So we would need to create a mock IMDSv2 service and then write the test cases. I have not found one test for the previous IMDSv1 call (https://github.com/paws-r/paws/blob/main/paws.common/tests/testthat/test_config.R). Can you point me to one?
> 
> Otherwise I propose to create a new issue for the integration test of IMDS (for both versions) as this probably is a bigger effort than providing the feature. Nevertheless, maybe you have already integration tests for other HTTP services used in paws?
> 
> For now I did only test manually the integration using AWS Sagemaker Notebooks with IMDSv1 enforced. The library without the change does not work there. Only after adding IMDSv2 supprot it works.
> 
> Let me know.

Have you considered a mock test? so we can stud the api call? I am happy to look into it if you haven't done mock testing before :)

jornfranke:
no problem to create a mocktest. Do you have a preferred framework or so? Then I can use the preferred framework.

I did not want create too many changes in paws as I am new to the paws code.

DyfanJones:
We haven't chosen a mock testing framework. I am accustomed with such frameworks as [mockery](https://github.com/r-lib/mockery) and [mockthat](https://github.com/nbenn/mockthat/). I reckon one of these should be fine :) Which ever one of these is the easiest to stub the api call should do :D We can always revisit mocking frameworks in the future :D

jornfranke:
Ok, I will try with mockery (for the reason that it seems to be better more frequently maintained and have more contributor). I have to think about where I put the mock (probably I will mock the function new_http_request) and integrate mockery in paws. It will take a bit of time, but should not be too much.

DyfanJones:
no worries :) if you need any help please feel free to reach out :)

DyfanJones:
@jornfranke the first thing i should of asked is. Is this done in the other aws sdk? :P I went straight into dev mode without asking (my bad hahaha).

jornfranke:
> @jornfranke the first thing i should of asked is. Is this done in the other aws sdk? :P I went straight into dev mode without asking (my bad hahaha).

not sure what you mean by "is this done in the other aws sdk"?

jornfranke:
I incorporated your comments. I also added two tests: one for the original functionality (that I did not implement) for IMDSv1 and one for the additional functionality (IMDSv2) that I implemented. Please let me know your feedback

jornfranke:
Btw.: Here you find how they do it in the official AWS Python SDK: https://github.com/boto/botocore/blob/master/botocore/utils.py#L372

I think it is roughly comparable how it is done for what I have implemented

paws.common/R/config.R
```diff
@@ -182,14 +182,53 @@ get_instance_metadata <- function(query_path = "") {
   if (trimws(tolower(get_env("AWS_EC2_METADATA_DISABLED"))) %in% c("true", "1")) {
     return(NULL)
   }
-
+  # Get token timeout for IMDSv2 tokens
+  tokentimeout=trimws(tolower(get_env("PAWS_EC2_IMDSV2_TOKEN_TIMEOUT_SECONDS")))
+  if (is.na(as.numeric(tokentimeout))) {
+        tokentimeout="30"
```
DyfanJones, line 188: Shouldn't we try and mimic what is being done in botocore? and use  `'21600'` as  our token timeout? https://github.com/boto/botocore/blob/master/botocore/utils.py#L376

paws.common/R/config.R
```diff
@@ -182,14 +182,41 @@ get_instance_metadata <- function(query_path = "") {
   if (trimws(tolower(get_env("AWS_EC2_METADATA_DISABLED"))) %in% c("true", "1")) {
     return(NULL)
   }
+  # Get token timeout for IMDSv2 tokens
+  tokentimeout=trimws(tolower(get_env("PAWS_EC2_IMDSV2_TOKEN_TIMEOUT_SECONDS")))
```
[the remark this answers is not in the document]
DyfanJones: I understand this gives us extra flexibility to the user however it seems `botocore` is using a private attribute: https://github.com/boto/botocore/blob/master/botocore/utils.py#L376. Should we do the same? :)

paws.common/R/config.R
```diff
@@ -182,14 +182,53 @@ get_instance_metadata <- function(query_path = "") {
   if (trimws(tolower(get_env("AWS_EC2_METADATA_DISABLED"))) %in% c("true", "1")) {
     return(NULL)
   }
-
+  # Get token timeout for IMDSv2 tokens
+  tokentimeout=trimws(tolower(get_env("PAWS_EC2_IMDSV2_TOKEN_TIMEOUT_SECONDS")))
+  if (is.na(as.numeric(tokentimeout))) {
```
DyfanJones, line 187: NIT: this can be simplified to `!nzchar(tokentimeout)`

paws.common/R/config.R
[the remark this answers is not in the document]
jornfranke: yes we can do the same, maybe we should not be more clever than AWS :) I will change it

paws.common/R/config.R
jornfranke: will be anyway removed with the change above

jornfranke:
thanks for the review so far. I have included the latest proposals

paws.common/R/config.R
```diff
@@ -177,19 +177,54 @@ get_profile_name <- function(profile = "") {
...
   )
+  if (token!="") {
+  metadata_request <-
+    new_http_request(
+        "GET", 
+        metadata_url, 
+        timeout = 1, 
+        header=c("X-aws-ec2-metadata-token"= token)
+    )
+  } else { # use IMDSv1 in case IMDSv2 is not available - not recommended - very insecure
```
davidkretch, line 224: is there a standard way in the AWS SDKs to force use of IMDSv2 (or v1), rather than defaulting to v1? this comment suggests that defaulting to v1 is a bad practice. if not, no change needed.

nit: I would make a separate get_metadata_request_imds_v1 and get_metadata_request_imds_v2 to make it easier to follow this logic.

davidkretch:
looks good to me, just had one question about forcing use of v2 only.

paws.common/R/config.R
jornfranke: As far as I know one can only configure it on the resource itself (EC2, Sagemaker Notebook etc.) to use IMDSv2-only, so you cannot just do it from the API itself.

Hence, I think splitting it up into get_metadata_request_imds_v1 and get_metadata_request_imds_v2 would not be really possible. I put them into one, because IMDSv2 is in essence just an additional HTTP request beforehand.

We could have a function fetch_metadata_token to fetch the token and imds_request that adds the token if is available or not (see also: https://github.com/boto/botocore/blob/master/botocore/utils.py#L569). 

Currently, I tried to integrate it into the existing logic in PAWS without rewriting what was already defined before.

davidkretch: Ok sounds good but in that case I’d get rid of “not recommended - very insecure” because someone who doesn’t know that background could delete this code inappropriately.

jornfranke: ok makes sense what you say. Maybe I will put security recommendations better in the documentation - the chance that somebody reads the documentation are higher ;)

davidkretch: Eh, that's not necessary.

jornfranke: ok removed the comment and added a reference to the AWS documentation on IMDS

DyfanJones:
@davidkretch is there anything else required? I think we are in a good place :)

davidkretch:
@DyfanJones  looks good to me :-)

jornfranke:
I saw that after the merge one of the pipelines failed (the two others went through):
https://github.com/paws-r/paws/actions/runs/3486467796/jobs/5832980669

It does not look directly related to the added code to me, but let me know if I can support troubleshooting

DyfanJones:
@jornfranke not to worry, retry of the failed job seemed to fix it. Thank you again for pr. Please feel free to pr in the future. We always are grateful for all prs as they really do help in the maintanence and development of paws :)

PR #578, opened 2023-01-01
Fix links to subcategory packages

davidkretch:
Links to functions in subcategory packages trigger warnings in CRAN check. This fixes the warnings by adding the package names the functions come from.


davidkretch:
Hello @DyfanJones, let me know what you think. We need to this to fix warnings that result from the introduction of subcategory packages.

DyfanJones:
Looks good to me :)

PR #581, opened 2023-01-07
Regenerate Paws for v0.2.0

davidkretch:
Regenerate Paws and update the version numbers of the category packages to 0.2.0.

Note: The main paws package can't be updated to 0.2.0 until the category packages are on CRAN because otherwise it'll fail the CI checks since its dependencies aren't available yet (because they're not on CRAN).


davidkretch:
@DyfanJones let me know what you think.

DyfanJones:
Looks all good to me :) The only change that will need to be done is paws.storage will need to restrict paws.common version to >=  0.5.4 to pick up the latest signer for `generate_presigned_url`.

PR #595, opened 2023-02-02
Fix invalid characters and DESCRIPTION

davidkretch:
* Add code to fix invalid characters that cause LaTeX problems.
* Fix the DESCRIPTION.
* Regenerate Paws.

### DIFF
--- a/cran/paws.application.integration/R/locationservice_operations.R
+++ b/cran/paws.application.integration/R/locationservice_operations.R
@@ -1845,9 +1845,7 @@ locationservice_put_geofence <- function(CollectionName, GeofenceId, Geometry) {
 #'
 #' @param IndexName &#91;required&#93; The name of the place index resource you want to use for the search.
 #' @param Language The preferred language used to return results. The value must be a valid
-#' [BCP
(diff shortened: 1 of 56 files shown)


davidkretch:
@DyfanJones this is ready -- the unicode characters causing problems are now gone. They are actually greek letters; currently make.paws will delete them, but long term I want to translate them to something else, although I'm not sure what if we can't use greek letters.

DyfanJones:
> @DyfanJones this is ready -- the unicode characters causing problems are now gone. They are actually greek letters; currently make.paws will delete them, but long term I want to translate them to something else, although I'm not sure what if we can't use greek letters.

Agreed. I believe R has some translation for them (https://cran.r-project.org/doc/manuals/R-exts.html#Mathematics). However not sure how to do this automatically, currently :S

davidkretch:
> > @DyfanJones this is ready -- the unicode characters causing problems are now gone. They are actually greek letters; currently make.paws will delete them, but long term I want to translate them to something else, although I'm not sure what if we can't use greek letters.
> 
> Agreed. I believe R has some translation for them (https://cran.r-project.org/doc/manuals/R-exts.html#Mathematics). However not sure how to do this automatically, currently :S

Ah good find.

PR #596, opened 2023-02-06
Set up paws sdk for 0.2.0 release

DyfanJones:


DyfanJones:
- Passing all Rhub tests with only size notes.
- Added tarball package size to each cran comment as "maintainer notes" for above rhub note.
- Removed any url links that causing status 404 or redirects
- Unescape any special escape characters

Issue #629, opened 2023-06-22
Cran html <span> raising note


DyfanJones:
Cran issue

```
Check: HTML version of manual, Result: NOTE
  Found the following HTML validation problems:
  serverlessapplicationrepository_cre_clo_for_cha_set.html:100:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:66): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:107:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:71): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:114:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:76): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:121:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:81): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:132:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:88): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:139:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:93): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:152:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:102): Warning: <span> proprietary attribute "href"
  serverlessapplicationrepository_cre_clo_for_cha_set.html:159:16 (serverlessapplicationrepository_cre_clo_for_cha_set.Rd:107): Warning: <span> proprietary attribute "href"
```

PR #630, opened 2023-06-22
fixes html span

DyfanJones:
Address: #629


DyfanJones:
Short term will add to end of `make_sdk` as it only affects 50 odd files. 

TODO: Unit tests

PR #636, opened 2023-06-26
Issue 631/Support for STS regional endpoint config

### LINKED ISSUE #631: `sts` endpoint - respect `sts_regional_endpoints = regional` or `AWS_STS_REGIONAL_ENDPOINTS`
Hi,
Thanks for developing and mainting `paws` packages. It seems like AWS config `sts_regional_endpoints = regional` and/or envvar `AWS_STS_REGIONAL_ENDPOINTS` are not respected by `paws`. Returned endpoint is not of form 
```
https://sts.{region}.amazonaws.com
```

but default
```
https://sts.amazonaws.com
```

This should be (hopefully) a simple fix. I can try to create PR, but would need some guidance on where exactly in the code base implement this. Thanks a lot!

joakibo:
Attempts to fix #631 . No testing done, needs input from R experts on the general strategy and placement of code.

* Remove hardcoded region for STS service when used in backend
* Add functions in config.R which will check the STS Regional Endpoints envvar + config file
* Dynamic endpoints list for the STS service, metadata endpoints is updated when STS client is initiated


cran/paws.security.identity/R/sts_service.R
```diff
@@ -102,6 +102,7 @@ NULL
 #' @rdname sts
 #' @export
 sts <- function(config = list()) {
+  .sts$metadata$endpoints <- .get_sts_endpoints(config$profile)
```
joakibo, line 105: This felt a little dirty, but since `endpoints` is dependent on `profile` this seemed like the natural way to do it.

DyfanJones:
I think I am understanding this issue a little better, I believe we need to create `custom_sts.R` script similar to https://github.com/paws-r/paws/blob/main/paws.common/R/custom_s3.R so that only `sts` methods are affect.  Will have a better look at this tomorrow :)

joakibo:
Sounds good, thanks. Having it in this custom script makes sense. I don’t have a good understanding of the code base, easier for you to deduce where things should go 👍

Changes outlined here should at least be sufficient as I see it.

DyfanJones:
@joakibo I believe we are in a good place with this. Will need to create some unit tests to fully test out the functionality. But I think it is nearly done. Please feel free to test it out and let me know

joakibo:
Thanks a lot @DyfanJones, looking forward to apply this fix on our side. Sorry for not being able to test this and assist more.

Issue #30, opened 2018-11-17
Add paginators


davidkretch:

edgBR:
Hi colleagues,

It seems that I face the paginator challenge myself :)

Was trying to get my all time historical trainings:

sm_client <- paws::sagemaker(config = list(region = myregion')
                             )

```r
total_training_jobs <- list()
j <- 1
sequence_var <- seq.POSIXt(from = as.POSIXct("2020-04-01 00:00:00"), to=as.POSIXct("2020-11-20 00:00:00"), by="hour")
for(i in sequence_var){
total_training_jobs[[j]] <- sm_client$list_training_jobs(MaxResults=100, CreationTimeAfter = i)
j <- j+1
}  
```

And I got a nice 400 ThrottlingException.

Anyone that has tried a workaround?

BR
/E

davidkretch:
Hey, sorry about that. I'll look into this this weekend. To my knowledge the approach to this is to delay requests some amount.

davidkretch:
I put together this attempt at a paginator. You supply it with your AWS API call as the argument to parameter `f` and it will take care of fetching each page of results and returning them as a list. Below this function is an example call. Let me know if this helps or not.

```r
# Get all pages of a given API call, retrying with exponential backoff.
paginate <- function(f, max_retries = 5) {
  resp <- f
  result <- list(resp)
  while ("NextToken" %in% names(resp) && length(resp$NextToken) > 0 && resp$NextToken != "") {
    next_token <- resp$NextToken
    call <- substitute(f)
    call$NextToken <- next_token
    # Retry with exponential backoff.
    # See https://docs.aws.amazon.com/general/latest/gr/api-retries.html.
    # See also https://github.com/paws-r/paws/blob/main/examples/error_handling.R.
    retry <- TRUE
    retries <- 0
    while (retry && retries < max_retries) {
      resp <- tryCatch(eval(call), error = function(e) e)
      if (inherits(resp, "error")) {
        if (retries == max_retries) stop(resp)
        wait_time <- 2^retries / 10
        Sys.sleep(wait_time)
        retries <- retries + 1
      }
      else retry <- FALSE
    }
    result <- c(result, list(resp))
  }
  return(result)
}
```

[...]

PR #650, opened 2023-07-24
Initial paginate function

DyfanJones:
This PR adds the following features:

- `paginate` function based off ticket: #30 
- re-export paws.common helper functions so that paws.common is not explicitly needed to be called

```r
library(paws.storage)

svc <- s3(confg())

paginate(
    svc$list_objects_v2(Bucket = "made-up")
)
```


DyfanJones:
I believe we are at a good point. All paginate information is passed from json file to paws so that we can build paginate queries correctly instead of guessing them. Next step will be to add unit tests.

edgBR:
Hi,

I am no longer using paws but wow, this is a great PR

PR #678, opened 2023-09-29
docs: `endpoint` & `region` in config, not credentials

dpprdan:
IIUC `endpoint` and `region` ought to items of `config`, not `config.credentials`


DyfanJones:
Ah thanks for the PR. I really appreciate it :) paws is a big beast and keep on top can be difficult at times :) I will review this when I get back from my holiday :)

PR #685, opened 2023-10-05
Service endpoint

DyfanJones:
Remove comments before for loop. Addresses issue highlighted in #667

### DIFF
--- a/paws.common/R/iniutil.R
+++ b/paws.common/R/iniutil.R
@@ -18,28 +18,27 @@ read_ini <- function(file_name) {
     scan(file_name, what = "", sep = "\n", quiet = T),
     perl = TRUE
   )
-  content_length <- length(content)
-
   # An empty credentials file is valid when using SSO
   # In that case, length(content) is 0.  Don't loop
   # in such a case, since 'grepl(..., content[i])'
   # will return logical(0), causing the 'if' to error out
-  if (content_length == 0) return(list())
+  if (length(content) == 0) return(list())
 
   # Get the profile name from an ini file
-  found <- grep("^\\[.*\\]", content)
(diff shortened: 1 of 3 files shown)


paws.common/R/iniutil.R
```diff
@@ -18,28 +18,27 @@ read_ini <- function(file_name) {
...
   # An empty credentials file is valid when using SSO
   # In that case, length(content) is 0.  Don't loop
   # in such a case, since 'grepl(..., content[i])'
   # will return logical(0), causing the 'if' to error out
-  if (content_length == 0) return(list())
+  if (length(content) == 0) return(list())
 
   # Get the profile name from an ini file
-  found <- grep("^\\[.*\\]", content)
   rm_els <- grep("(^;)|(^#)", content)
```
dpprdan, line 28: You may want to allow for indented comments, see [this beauty of a config file](https://www.scaleway.com/en/docs/storage/object/api-cli/object-storage-aws-cli/#how-to-install-the-aws-cli) in the wild.

```suggestion
  rm_els <- grep("(^;)|(^\\s+#)", content)
```

DyfanJones: I will add it to the test data_ini to make sure it removes them correctly :)

DyfanJones:
I believe the new test ini file should cater for most scenarios https://github.com/paws-r/paws/blob/7c10b6bfeb99e30974748c36ded79ddc9967ba2b/paws.common/tests/testthat/data_ini

When I looked into the package `ini` it didn't handle nested content at all and it didn't could falter on some comments.

The bright side the new `read_ini` seems to run better than the old method and the ini package.
image

PR #690, opened 2023-10-16
remove repeat processes

DyfanJones:
I believe this is the fastest I can get `read_ini` function without going full c++.

We have some decent performance now with roughly ~500us to read in the test data_ini compared to ~3ms. This should help with any extra calls to the config files we need to make.

image


DyfanJones:
It seems like we are getting some good performance compared to the `M.ini` package https://github.com/DyfanJones/M.ini (a simple rcpp wrapper around the mINI header only C++ library).

image

PR #692, opened 2023-10-17
cache read_ini results

DyfanJones:


DyfanJones:
Just for completeness, as expected caching drastically improves the performance.


image

I am really happy with this, as we can include more service setting from the .aws/config (https://github.com/paws-r/paws/issues/667) without affecting the overall performance.

PR #695, opened 2023-10-25
Add support for web token based IAM in config files

liuquinlin:
Adding support for web_identity_token_file in the AWS config file (see https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc) - also opportunistically adding some unit testing in that area.

I need this functionality for a project I'm working on that uses this auth method, let me know if there's anything else I need to do to enable this change (I've successfully tested this for my use case by modifying the source via `trace`).

### DIFF
--- a/paws.common/DESCRIPTION
+++ b/paws.common/DESCRIPTION
@@ -32,6 +32,7 @@ Suggests:
     covr,
     crayon,
     mockery,
+    withr,
     rstudioapi,
     testthat
 SystemRequirements: pandoc (>= 1.12.3) - http://pandoc.org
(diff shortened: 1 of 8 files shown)


DyfanJones:
Hi @liuquinlin  thanks for raising a PR. I really appreciate the contribution. I will look it over :)

DyfanJones:
It looks all good to me, the only issue i am getting is:

```r
Error (test_credential_providers.R:114:5): config_file_provider
Error in `eval(value, envir = return_values_env)`: object 'profile_creds' not found
```
When I run the tests locally

liuquinlin:
Thanks for the review! Yeah sorry about that think that was a mistaken leftover from a rename

PR #700, opened 2023-11-02
new aws services

DyfanJones:
This PR adds the latest AWS services to paws.

@davidkretch @adambanker  do you mind having a quick look over the services. Just making sure I have added them to the correct category package


make.paws/inst/extdata/packages.yml
```diff
@@ -318,6 +347,7 @@
     - cognito-identity
     - cognito-idp
     - cognito-sync
+    - codeguru-security
```
davidkretch, line 350: I think this should go in developer tools with the other code guru services, since it is for analyzing code and not infra.

make.paws/inst/extdata/packages.yml
```diff
@@ -378,11 +411,15 @@
...
     - mwaa
+    - pipes
+    - resource-explorer-2
     - schemas
+    - scheduler
     - states
     - sns
     - sqs
     - swf
+    - simspaceweaver
```
davidkretch, line 422: I think belongs in the gaming package.

make.paws/inst/extdata/packages.yml
```diff
@@ -412,16 +451,19 @@
...
+    - email
+    - entityresolution
     - wisdom
     - customer-profiles
     - pinpoint
     - pinpoint-email
-    - sms-voice
     - pinpoint-sms-voice-v2
-    - email
+    - medical-imaging
```
davidkretch, line 464: I think this should go in business applications.

make.paws/inst/extdata/packages.yml
```diff
@@ -412,16 +451,19 @@
   services:
     - connect
     - connectcampaigns
+    - connectcases
     - connect-contact-lens
     - connectparticipant
+    - email
+    - entityresolution
```
davidkretch, line 458: This one is tough. Maybe in analytics since it is for determining which database records are really referring to the same thing. I don’t think it belongs in customer engagement though.

davidkretch (commented):
Overall looks good.

make.paws/inst/extdata/packages.yml
DyfanJones: I found this one hard as well. I initially thought customer due to the description.

> With Entity Resolution, you can match source records containing consumer identifiers, such as name, email address, and phone number. This is true even when these records have incomplete or conflicting identifiers. 

But more that happy to move it over to analytics

PR #707, opened 2023-11-09
Simplify `set_os_env_cache()` cache to work on Posit Workbench

fh-mthomson:
Closes #706 

Introduced in https://github.com/paws-r/paws/pull/693

PR #708, opened 2023-11-09
Update to testthat 3e

fh-mthomson:
Ran `usethis::use_testthat(3)` ([reference](https://testthat.r-lib.org/articles/third-edition.html)) and incorporated suggested tweaks to tests to reflect testthat deprecations.

### DIFF
--- a/make.paws/DESCRIPTION
+++ b/make.paws/DESCRIPTION
@@ -36,7 +36,7 @@ Imports:
 Suggests: 
     covr,
     formatR,
-    testthat (>= 3.0.0),
+    testthat,
     paws.storage,
     withr,
     mockery
@@ -67,5 +67,3 @@ Collate:
     'sdk_helper.R'
     'cran_sub_category.R'
     'paws_helper_fun.R'
-Config/testthat/edition: 3
-Config/testthat/parallel: true
--- a/make.paws/DESCRIPTION
+++ b/make.paws/DESCRIPTION
@@ -36,7 +36,7 @@ Imports:
 Suggests: 
     covr,
     formatR,
-    testthat,
+    testthat (>= 3.0.0),
     paws.storage,
     withr,
     mockery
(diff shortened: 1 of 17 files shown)


fh-mthomson:
@DyfanJones I'm scratching my head over why this fails on the CI, when it runs on Workbench and Mac with no errors:
```
── R CMD check results ───────────────────────────────── paws.common 0.6.3.9000 ────
Duration: 1m 19.4s

0 errors ✔ | 0 warnings ✔ | 0 notes ✔

R CMD check succeeded
```

Specifically, the error is consistently:
```
Error ('test_custom_s3.R:382:3'): redirect error without region ─────────────
Error in `get_region(cfgs[["credentials"]][["profile"]])`: No region provided
```

Of note, it includes all the changes in #707, so I'm wondering if there's a subtle bug in that branch such that, when using `testthat` 3e, with slightly more rigid standards, something isn't up to snuff?

DyfanJones:
I don't think there is a bug in #707 as it passed all the tests 🤔 will have a look at this PR now

DyfanJones:
I am not sure what `Config/testthat/parallel: true` would do. I guess if it fails again we can turn that part off to see if that is causing the weirdness 🤔

fh-mthomson:
> I am not sure what `Config/testthat/parallel: true` would do. I guess if it fails again we can turn that part off to see if that is causing the weirdness 🤔

It should only impact how `devtools::test()`, not `devtools::check()` runs. 

That said, it's possible that it's exposing some underlying impurities in how tests are run. Let's say there are two test files (A, B) run in parallel. If test A sets an env var that isn't unset, and A runs before B, then B would be indirectly affected by the env var set in A. `withr::local_environment()` would be a longer-term mitigation here.

fh-mthomson:
Well, that confirms it!

PR #972, opened 2026-04-30
No content-length for empty-body GET/HEAD

noamross:
This PR makes a small change to the V4 signing method to make it compatible with services that implement it strictly, in my case Cloudflare R2.  For requests with empty bodies, the signature fails if the body is empty with the default "empty SHA". If instead, we remove the content-length header, this succeeds.


DyfanJones:
Hi @noamross, this looks all good to me. Thank you for your contribution.

PR #992, opened 2026-06-22
feat: add S3 on Outposts support (signing_name + content-sha256)

mateuslsl:
Problem: paws doesn't support S3 on Outposts Access Points. Requests fail with AuthorizationHeaderMalformed (signing service mismatch) and Missing x-amz-content-sha256 header.

Solution: Add "s3-outposts" to the service name checks in the SigV4 signer and body digest handler, so Outposts endpoints receive the correct credential scope and required headers.

### DIFF
--- a/paws.common/R/signer_v4.R
+++ b/paws.common/R/signer_v4.R
@@ -125,11 +125,6 @@ sign_sdk_request_with_curr_time <- function(request, curr_time_fn = now, opts = 
     name <- request$config$service_name
   }
 
-  # Auto-detect S3 on Outposts from endpoint URL
-  if (name == "s3" && grepl("s3-outposts", request$config$endpoint %||% "")) {
-    name <- "s3-outposts"
-  }
-
   v4 <- Signer(
     credentials = request$config$credentials,
     disable_header_hoisting = request$not_hoist,
@@ -137,7 +132,7 @@ sign_sdk_request_with_curr_time <- function(request, curr_time_fn = now, opts = 
     disable_request_body_overwrite = TRUE
   )
 
-  if (name %in% c("s3", "s3-outposts")) {
+  if (name == "s3") {
     v4$disable_uri_path_escaping <- TRUE
   }
 
@@ -374,8 +369,8 @@ build_body_digest <- function(ctx) {
   hash <- get_element(ctx$request$header, "X-Amz-Content-Sha256")
   if (hash == "") {
     include_sha256_header <- (ctx$unsigned_payload ||
-      ctx$service_name %in% c("s3", "s3-object-lambda", "glacier", "s3-outposts"))
-    s3_presign <- (ctx$is_presigned && ctx$service_name %in% c("s3", "s3-object-lambda", "s3-outposts"))
+      ctx$service_name %in% c("s3", "s3-object-lambda", "glacier"))
+    s3_presign <- (ctx$is_presigned && ctx$service_name %in% c("s3", "s3-object-lambda"))
     if (ctx$unsigned_payload || s3_presign) {
       hash <- "UNSIGNED-PAYLOAD"
       include_sha256_header <- !s3_presign
--- a/paws.common/R/signer_v4.R
+++ b/paws.common/R/signer_v4.R
@@ -132,7 +132,7 @@ sign_sdk_request_with_curr_time <- function(request, curr_time_fn = now, opts = 
     disable_request_body_overwrite = TRUE
   )
 
-  if (name == "s3") {
+  if (name %in% c("s3", "s3-outposts")) {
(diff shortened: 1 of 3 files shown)


paws.common/R/signer_v4.R
```diff
@@ -125,14 +125,19 @@ sign_sdk_request_with_curr_time <- function(request, curr_time_fn = now, opts =
     name <- request$config$service_name
   }
 
+  # Auto-detect S3 on Outposts from endpoint URL
+  if (name == "s3" && grepl("s3-outposts", request$config$endpoint %||% "")) {
+    name <- "s3-outposts"
+  }
+
```
DyfanJones: I don't think we need to detect the endpoint to overload the signing name. The signing-name/service name should be s3-output for s3-output endpoints:

https://github.com/paws-r/paws/blob/main/cran/paws.storage/R/s3outposts_service.R

paws.common/R/signer_v4.R
```diff
@@ -125,14 +125,19 @@ sign_sdk_request_with_curr_time <- function(request, curr_time_fn = now, opts =
...
   v4 <- Signer(
     credentials = request$config$credentials,
     disable_header_hoisting = request$not_hoist,
     curr_time_fn = curr_time_fn,
     disable_request_body_overwrite = TRUE
   )
 
-  if (name == "s3") {
+  if (name %in% c("s3", "s3-outposts")) {
```
DyfanJones: Is this true? When I look at aws sdk go it only does s3: https://github.com/aws/aws-sdk-go/blob/070853e88d22854d2355c2543d0958a5f76ad407/aws/signer/v4/v4.go#L472

Can you reference another aws sdk that does it with s3-outpost

paws.common/R/signer_v4.R
```diff
@@ -369,8 +374,8 @@ build_body_digest <- function(ctx) {
   hash <- get_element(ctx$request$header, "X-Amz-Content-Sha256")
   if (hash == "") {
     include_sha256_header <- (ctx$unsigned_payload ||
-      ctx$service_name %in% c("s3", "s3-object-lambda", "glacier"))
-    s3_presign <- (ctx$is_presigned && ctx$service_name %in% c("s3", "s3-object-lambda"))
+      ctx$service_name %in% c("s3", "s3-object-lambda", "glacier", "s3-outposts"))
+    s3_presign <- (ctx$is_presigned && ctx$service_name %in% c("s3", "s3-object-lambda", "s3-outposts"))
```
DyfanJones: Possibly worth double checking if we have all services:

https://github.com/aws/aws-sdk-go/blob/070853e88d22854d2355c2543d0958a5f76ad407/aws/signer/v4/v4.go#L700C3-L728

DyfanJones (commented):
Just one or two questions around implementation to ensure we are capturing everything

paws.common/R/signer_v4.R
mateuslsl: You're right — I've reverted this. I was aware of `paws::s3outposts()`, but that service only exposes the control plane (`CreateEndpoint`, `DeleteEndpoint`, `ListEndpoints`), not S3 data operations like `list_objects_v2`, `get_object`, etc.

My use case is accessing S3 data on Outposts via `paws::s3()` with a custom endpoint. In aws-sdk-go-v2, this is handled by [ARN-parsing middleware](https://github.com/aws/aws-sdk-go-v2/tree/main/service/s3/internal/arn) that overrides the signing name automatically. Since paws doesn't have ARN resolution, I was working around it.

Would a configurable `signing_name` in the client config be a cleaner approach?

paws.common/R/signer_v4.R
mateuslsl: You're correct — the Go SDK only disables URI path escaping for `"s3"`. I couldn't find another AWS SDK that extends this to `s3-outposts`. Reverted in the latest commit.

paws.common/R/signer_v4.R
mateuslsl: Good call. Checked the Go SDK reference — the list is `s3`, `s3-object-lambda`, `glacier` for `include_sha256_header`, and `s3`, `s3-object-lambda` for presign. No `s3-outposts`. Reverted to match.

DyfanJones:
Apologies in my slow response to this. It has been abit crazy with family and kids. Is it that simple with an override in the signing name? Or should we have a middleware similar to aws sdk go v2?

DyfanJones:
Apologies in the delay. I have been abit busy with family and work. Sadly I have a family holiday coming up and won't be able to get to this until I am back. If it asap please let me know and I will do my best to get it into r-universe and that

Issue #995, opened 2026-07-31
s3$delete_object() 403 erros with Cloudflare


Aariq:
I think possibly @noamross's change in #988 (linked below) also needs to apply to DELETE requests.  I'm getting errors like:
```
Error:
! SignatureDoesNotMatch (HTTP 403). The request signature we calculated does not match the signature you provided. Check your secret access key and signing method.

Hide Traceback
    ▆
 1. └─s3$delete_object(Bucket = bucket, Key = key) at untitled:Untitled-1:59:3
 2.   └─paws.common::send_request(request)
 3.     └─paws.common:::retry(request)
 4.       └─paws.common:::run(request, retry)
 5.         └─handler$fn(request)
 ```

https://github.com/paws-r/paws/blob/ef2015286fa0f1596287ecefaa53d8c6908e7954/paws.common/R/signer_v4.R#L395-L396

PR #996, opened 2026-07-31
don't sign empty-body DELETE requests either

Aariq:
Fix for #995.  I've tested this out with a Cloudflare R2 bucket and deleting objects now works.


DyfanJones:
Apologies in the late reply. I have been on holiday and been fixing my other r packages.

DyfanJones:
Apologies for the delay again. Been fairly busy with work. I should have this in shortly.

